PR #5, opened 2023-02-04
fix CI 

jerryjliu:
pip install latest version of gpt-index that has download_loader


jerryjliu:
hmm still can't see CI - is this a permissions thing? 

i'll revisit this tmrw morning, going to head off to bed

emptycrown (approved):

PR #12, opened 2023-02-08
Unstructured.io loader

emptycrown:
New loader for unstructured.io. README has all the details!


loader_hub/file/unstructured/README.md
```diff
@@ -0,0 +1,38 @@
...
+```python
+from pathlib import Path
+from gpt_index import download_loader
+
+SimpleDirectoryReader = download_loader("SimpleDirectoryReader")
+
+loader = SimpleDirectoryReader()
+documents = loader.load_data('./data', file_extractor={
+  ".pdf": "UnstructuredReader",
+  ".html": "UnstructuredReader",
```
jerryjliu, line 32: oh wow nice

loader_hub/file/unstructured/base.py
```diff
@@ -0,0 +1,35 @@
...
+        # Prerequisite for Unstructured.io to work
+        import nltk
+
+        nltk.download("averaged_perceptron_tagger")
+
+    def load_data(
+        self, file: Path, extra_info: Optional[Dict] = None
+    ) -> List[Document]:
+        """Parse file."""
+        from unstructured.partition.auto import partition
```
jerryjliu, line 30: we could probably start not doing lazy imports since we have a requirements.txt file now

jerryjliu (approved):
sweet!

loader_hub/file/unstructured/base.py
```diff
@@ -0,0 +1,35 @@
...
+
+        elements = partition(str(file))
+        text = "\n\n".join([" ".join(str(el).split()) for el in elements])
```
jerryjliu, line 33: there could be an option to combine partitions into one Document or separate Documents - with the latter, we can basically use unstructured capabilities for smart text splitting

jerryjliu: actually we could clarify this in the README too, as unstructured helps to solve some of the pain points with naive text splitting that gpt index currently implies. By default gpt index splits a Document only if it needs to (if it's above the specified chunk size). By "pre-splitting into multiple docs here, we basically offer users an alternative, smarter form of text splitting

loader_hub/file/unstructured/base.py
emptycrown: IMO still better to do lazy. Otherwise, for example, loading SimpleDirectoryReader would import a million things. Having the requirements txt at least eliminates the need to do try/except to check if things are installed. What do you think?

emptycrown: Actually I refactored SimpleDirectoryReader so that it doesn't do that anymore. But still feels safer to not be importing unnecessarily.

loader_hub/file/unstructured/base.py
emptycrown: Makes sense

cragwolfe: one simple approach to experiment with could be to create a new document for each `unstructured.documents.elements.Title` and include everything underneath until the next Title. obviously not a perfect solution, and would depend on the html quite a bit.

cragwolfe: at the very least though, one should rarely have to split an individual text element -- the exception being a really long `.NarrativeText` element.

emptycrown: Interesting, I'll note this as a todo to experiment with

loader_hub/file/unstructured/base.py
jerryjliu: oh ok. that's fair, yeah not a big deal for now

jerryjliu (approved):

PR #13, opened 2023-02-08
PR to add a loader that crawls and scrapes online knowledge bases

jasonwcfan:
Love to see a central repo for data loaders for LLMs. I've been working on similar stuff for web so thought I'd contribute some of it.


emptycrown (approved):

loader_hub/web/knowledgebase_web/base.py
```diff
@@ -0,0 +1,101 @@
+from typing import List
+import time
+from playwright.sync_api import sync_playwright
```
emptycrown, line 3: Loader-specific packages should be imported inside the class so that they're only imported when the class is instantiated or called. (Also, the tests won't pass otherwise because the package isn't installed)

loader_hub/web/knowledgebase_web/README.md
```diff
@@ -0,0 +1,94 @@
+# Knowledge Base Website Loader
+
+This loader is a web crawler and scraper that fetches text content from websites structed as knowledge bases. Typically these sites have a directory structure with several sections and many articles in each section. This loader crawls and finds all links that match the article path provided, and scrapes the content of each page.
```
emptycrown, line 3: Maybe it'd be good to add some examples in this first bit of what knowledge bases are? I wasn't super clear the first time reading through.

emptycrown (changes requested):
Left a couple of comments. Main thing is to import packages lazily when the class is called or initialized.

emptycrown (approved):

PR #14, opened 2023-02-08
Fix file loading in `PandasCSVReader`

AjinkyaIndulkar:
## Description

I found a typo in README to use `PandasCSVReader`.

**Edit**: Reverted changes to README and updated the class instead.

### Changelog

- 🐛 fixed file loading in `PandasCSVReader`

### DIFF
--- a/loader_hub/file/pandas_csv/base.py
+++ b/loader_hub/file/pandas_csv/base.py
@@ -52,7 +52,7 @@ class PandasCSVReader(BaseReader):
         self._row_joiner = row_joiner
         self._pandas_config = pandas_config
 
-    def parse_file(
+    def load_data(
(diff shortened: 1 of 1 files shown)


emptycrown:
Good catch! It's actually the class itself that's wrong haha (in `base.py`) and not the README. Feel free to make that change, or I will later.

AjinkyaIndulkar:
> Good catch! It's actually the class itself that's wrong haha (in `base.py`) and not the README. Feel free to make that change, or I will later.

I see! I thought it was by design. I'll make the change.

AjinkyaIndulkar:
@emptycrown can you please review again?

emptycrown (approved):

PR #18, opened 2023-02-09
S3 reader plus documentation

emptycrown:
S3 loader that can load a file or directory


loader_hub/web/knowledge_base/base.py
```diff
@@ -26,7 +28,15 @@ class KnowledgeBaseWebReader(BaseReader):
             e.g. '.article-body'
     """
 
-    def __init__(self, root_url, link_selectors, article_path, title_selector, subtitle_selector, body_selector) -> None:
+    def __init__(
+        self,
+        root_url,
```
jerryjliu, line 33: haven't noticed this before, but hm it'd be nice to have type annotations if possible

i have a todo to add a formatter/linter to llamahub that doesn't need to install every external dependency

jerryjliu (approved):

loader_hub/web/knowledge_base/base.py
emptycrown: sounds good -- I'll edit those separately

PR #25, opened 2023-02-10
Add google drive reader for  text files

ravi03071991:


emptycrown:
Oh nice! So does this only work for .txt files at the moment?

ravi03071991:
yeah.

jerryjliu (commented):
@ravi03071991 one idea here which is similar to what @emptycrown did in https://github.com/emptycrown/llama-hub/pull/17/files. first just download the files from google drive to a local, temporary directory, and then call `download_loader` on "SimpleDirectoryReader" to parse these files. That way, you get to parse all extensions that SimpleDirectoryReader is able to process!

ravi03071991:
> @ravi03071991 one idea here which is similar to what @emptycrown did in https://github.com/emptycrown/llama-hub/pull/17/files. first just download the files from google drive to a local, temporary directory, and then call `download_loader` on "SimpleDirectoryReader" to parse these files. That way, you get to parse all extensions that SimpleDirectoryReader is able to process!

Yeah, even I thought the same. So, as per the [documentation](https://developers.google.com/drive/api/guides/manage-downloads), google drive allows us to download files in bytes format and I am unable to figure out how to extract the text from bytes format. Any directions that you guys @jerryjliu @emptycrown can suggest here or should I try doing it in a different route?

ravi03071991:
@jerryjliu I found a workaround. Will update the PR with your suggesstion.

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # Get metadata of the file
+            metadata = None
+            if self.file_metadata is not None:
+                metadata = self.file_metadata(str(input_file))
+
+            document = ""
+
+            # If suffix in fileextractor, user reader
+            if suffix in self.file_extractor:
```
emptycrown, line 131: I feel like it makes more sense to call `SimpleDirectoryReader` here even if it's just a single file so that we don't need to duplicate things like `DEFAULT_FILE_EXTRACTOR`.

Alternatively, we could also make this function `_download_file_id` instead and call that from `_load_from_file_ids` to download all the files to a single temp directory. Then use `SimpleDirectoryReader`.

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,197 @@
...
+        try:
+
+            # Authenticate and build the Drive API client
+            service = build('drive', 'v3', credentials=creds)
+
+            # Get the file details
+            file = service.files().get(fileId=file_id).execute()
+
+            # Get file name that will be downloaded
+            input_file = file['name']
```
emptycrown, line 114: It doesn't look like these files are getting cleared out after they're downloaded. Perhaps `tempfile` would be a good option. Check out the example [here](https://github.com/emptycrown/llama-hub/blob/main/loader_hub/s3/base.py) that uses `tempfile.TemporaryDirectory()`. The files are deleted after the code is run.

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,197 @@
...
+            # Download the file locally
+            file_obj = drive.CreateFile({'id': file_id})
+            file_obj.GetContentFile(input_file)
+
+            # Get suffix to read the content
+            suffix = Path(input_file).suffix
+
+            # Get metadata of the file
+            metadata = None
+            if self.file_metadata is not None:
```
emptycrown, line 125: Is it possible to pull metadata from the Google API? Like author, date, etc.?

emptycrown (changes requested):
Thank you for making this! This looks awesome. I left a couple of comments; let me know what you think.

loader_hub/googledrive/base.py
ravi03071991: Agreed. Will make necessary changes.

loader_hub/googledrive/base.py
ravi03071991: Yeah. This is something that should be taken care of. Will delete files once they are downloaded.

loader_hub/googledrive/base.py
ravi03071991: But files can be of varied formats such as .ipynb files for which we don't want to read in a folder. That's the reason I explicitly used  `DEFAULT_FILE_EXTRACTOR` to include specific format files. What do you think on it?

loader_hub/googledrive/base.py
ravi03071991: id,name,createdTime,modifiedTime,size - these are the metadata available for a file. May be including name, createdTime, modifiedTime should be good idea as a part of metadata?

emptycrown: Yep agreed!

loader_hub/googledrive/base.py
emptycrown: If that's a behavior we want (e.g. skip a file if there's no explicit extractor for its suffix) then it should be added into `SimpleDirectoryReader` so that we can specify that with an arg, since that is not specific to just Google Drive.

In most cases, we don't want to do that anyway, since it will miss a lot of text-based files we could have parsed successfully, like .md for example. If the user knows that their data is in certain types of files, they can always use the `required_exts` argument in SimpleDirectoryReader.

emptycrown (commented):
If that's a behavior we want (e.g. skip a file if there's no explicit extractor for its suffix) then it should be added into SimpleDirectoryReader so that we can specify that with an arg, since that is not specific to just Google Drive.

In most cases, we don't want to do that anyway, since it will miss a lot of text-based files we could have parsed successfully, like .md for example. If the user knows that their data is in certain types of files, they can always use the required_exts argument in SimpleDirectoryReader.

loader_hub/googledrive/base.py
ravi03071991: Understood. Will update it.

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        try:
+            file = drive.CreateFile({'id': fileid})
+            # download file with filename
+            file.GetContentFile(filename)
+        except Exception as e:
+            logger.error(
+                'An error occurred while downloading file: {}'.format(e))
+
+    def _load_data_fileids_meta(self, fileids_meta: str) -> List[Document]:
```
emptycrown, line 134: Seems like `fileids_meta` should be `List[str]`?

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,220 @@
...
+    def __init__(self,
+                 credentials_path: str = "credentials.json",
+                 client_secrets_path: str = "client_secrets.json",
+                 token_path: str = "token.json") -> None:
+        """Initialize with parameters."""
+        self.credentials_path = credentials_path
+        self.client_secrets_path = client_secrets_path
+        self.token_path = token_path
+
+    def _get_credentials(self) -> Any:
```
emptycrown, line 39: I think there's something broken with this function. It's trying to refresh and make me login every single time I run it.

loader_hub/googledrive/README.md
```diff
@@ -0,0 +1,38 @@
...
+For example, the folder_id of https://drive.google.com/drive/folders/1w7XryYu6mL9VLmfyqUkA4_fRnDbsCqV- is 1w7XryYu6mL9VLmfyqUkA4_fRnDbsCqV-
+
+### file_id
+You can extract a file_id directly from its sharable drive URL. 
+
+For example, the file_id of https://drive.google.com/file/d/1LEqD_zQiOizKrBKZYKJtER_h6i49wE-y/view?usp=sharing is 1LEqD_zQiOizKrBKZYKJtER_h6i49wE-y
+
+## Usage
+
+We need `credentials.json` and `client_secrets.json` files to use this reader.
```
emptycrown, line 19: In addition to credentials, I also had to enable Google Drive API in my Google API Console for the app corresponding to these credentials. Let's make a note of that too?

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,220 @@
...
+            with tempfile.TemporaryDirectory() as temp_dir:
+                def get_metadata(filename):
+                    return metadata[filename]
+
+                temp_dir = Path(temp_dir)
+                metadata = {}
+
+                for fileid_meta in fileids_meta:
+                    filename = fileid_meta[2]
+                    filepath = os.path.join(temp_dir, filename)
```
emptycrown, line 152: Will there be issues here when there are nested files/directories?

emptycrown (changes requested):
Looks good overall but I'm having some trouble getting it to run smoothly locally. Do you have an .ipynb or something you can share?

loader_hub/googledrive/base.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def load_data(self, folder_id: str = None, file_ids: List[str] = None) -> List[Document]:
+        """Load data from the folder id and file ids.
+        Args:
+            folder_id: folder id of the folder in google drive.
+            file_ids: file ids of the files in google drive.
+        Returns:
+            List[Document]: A list of documents.
+        """
+        global creds, drive
```
emptycrown, line 214: Do these have to be globals? Can't we just do `self.creds` or something?

loader_hub/googledrive/base.py
ravi03071991: ohh yeah. Will update.

loader_hub/googledrive/base.py
ravi03071991: that's because of following part
`gauth = GoogleAuth()
  gauth.LocalWebserverAuth()
  drive = GoogleDrive(gauth)`

loader_hub/googledrive/README.md
ravi03071991: Yes, sure. Will do that.

emptycrown:
Helped clean up / refactor it a bit. Let me know what you think!

loader_hub/googledrive/base.py
ravi03071991: yes it works.

ravi03071991:
> Helped clean up / refactor it a bit. Let me know what you think!

Thank you. Was restructuring in a similar manner. I think now we are good to go.

emptycrown (approved):

PR #34, opened 2023-02-14
Add GitHub Repository Reader 

ahmetkca:
I have opened a PR against [gpt_index PR#439](https://github.com/jerryjliu/gpt_index/pull/439:) that would pair with this PR.

That PR allows me to use/import `GithubClient` from gpt_index's `download_loader`

Adds reader for GitHub repositories (`GithubRepositoryReader`)
- accepts `github_client`, (I am planning to make the client `GitClient` and add `BitBucket` and `GitLab` as well.
   - pass GitHub access token to the github_client
- accepts `owner`, `repo` at the constructor level.
- `use_parser` allows it to parse the supported files by utilizing existing `DEFAULT_FILE_EXTRACTOR` such as images, videos, markdown etc. files. Defaults to `false` and it requires all the dependencies of parsers (such as pytorch, sentencepeiece etc) if set to `true`
- optionally pass a list of file extensions to type the filter to allow or ignore file extensions i.e. `([".png", ".jpeg", ".gif"], GithubRepositoryReader.FilterType.EXCLUDE)`
- optionally pass a list of directories and type the filter to allow or ignore directories i.e. `(["gpt_index/indices/vector_store", "gpt_index/embeddings", GithubRepositoryReader.FilterType.INCLUDE)`
- it depends on `httpx` as I used it in the `GithubClient` instead of requests (because it has better support for async requests)
- `concurrent_requests` allows you to control the number of concurrent requests the github_client will make. The greater the number the more unreliable it is. (default: 5)

It recursively fetches tree objects and constructs their full path relative to root of the repository. After all the tree objects (blobs with tree_sha) fetched, they are passed to `BufferedAsyncIterator`. Only `buffer_size` of tree objects fetched concurrently and stored at once, thus less memory usage during the generation of documents. (I was initially looping over the tree objects and fetching blobs one at a time but as I was testing with a large repository it was consuming a lot of memory and it was slow. 
Generated documents contain decoded base64-encoded-string in utf-8 as `text` and file name and full path as `extra_info`

#### **Testing**

- Tested in FastAPI environment with multiple repositories at the same time.
- Tested with GPTxxxIndexes. To get the best result out of repository I found mixing tree, list and vector index gives better result but it is also more costly.


emptycrown:
@ahmetkca are you able to write a test for this loader?

ahmetkca:
> @ahmetkca are you able to write a test for this loader?

Btw is simple test enough or do you want proper test suite with mocks and stuff?

emptycrown:
> Btw is simple test enough or do you want proper test suite with mocks and stuff?

A simple test is perfectly fine if you have something in mind that can be done without network requests. If not, we can skip for now and just add an example ipynb into a new `examples` directory.

ahmetkca:
Hey @emptycrown, I finally got time to work on this. I added initial test which uses network request but I think we can ignore by default. If somebody wants to make change to `github_repo` they can specifically run this test and hopefully another one coming. The inital test doesn't take long tho I am only checking one deep which I think is enough for now.

PR #35, opened 2023-02-14
Add Japanese PDF reader

JiroShimaya:
I created a loader (JapanesePDFReader) to parse Japanese PDFs. Please review it.

# Changes
- Added JapanesePDFReader class. This class uses the [pdfminer.six](https://github.com/pdfminer/pdfminer.six) library to parse Japanese PDFs that the existing PDFReader cannot.
# Reasons for changes
- The existing PDFReader had a problem of garbling some Japanese PDFs. This is probably due to the character encoding, as the PyPDF2 library used in PDFReader does not support parsing characters other than utf8.
- As an example of a PDF that garbles, I give the following URL.
  - https://www8.cao.go.jp/chosei/koubun/about/shikumi/g_bun/kenshuu.pdf

Thank you for your cooperation.


emptycrown:
Thanks for the PR and looks good! My only comment is: is it correct to name this loader `JapanesePDFReader`? Wouldn't it work for many other languages?

emptycrown (approved):

JiroShimaya:
Thank you for the review. Sorry for the delay. Here is the intention of my last commit.

1.
I named it CJKPDFReader. CJK stands for Chinese, Japanese, and Korean languages. As you correctly pointed out, pdfminer.six can read languages other than Japanese, so I was not confident that JapanesePDFReader was an appropriate name, but I saw that pdfminer.six’s docs said “support CJK languages”, so I thought it would be suitable to use this as the class name. https://github.com/pdfminer/pdfminer.six#:~:text=CJK%20languages%20and%20vertical%20writing%20scripts%20support

2. 
I also improved the text extraction process. The previous code sometimes failed to extract text, but the current code works better. Here is an example of a PDF that the current code can parse, but the previous code could not. https://www.beret.co.jp/books/tachiyomi/images/563.pdf

emptycrown:
Got it -- thanks! I'll update the readme.

PR #36, opened 2023-02-15
Better tests, including download_loader

emptycrown:
Some more comprehensive library tests, plus some loader tests


tests/test_download
```diff
@@ -0,0 +1,12 @@
+"""Check that downloading loaders from GPT Index is working as expected."""
```
jerryjliu, line 1: is this supposed to be a py file?

tests/test_file.py
```diff
@@ -0,0 +1,207 @@
+"""Test file reader."""
+
+from tempfile import TemporaryDirectory
+from typing import Any, Dict
+
+from gpt_index import download_loader
```
jerryjliu, line 6: ideally unit tests shouldn't perform network requests, also this isn't technically testing the "current" code that's within the current branch of llama-hub but rather the code in master, since `download_loader` pulls from master.

may be worth investigating using a mocked `download_loader` function which simply does a lookup in library.json in the current branch instead of making a network request

jerryjliu (changes requested):
thanks! some comments

tests/test_file.py
emptycrown: Testing "current" code makes sense. What's typically the correct way to test cross-repo stuff like this though? The goal would be to catch issues like when there was the breaking `verbose` change?

jerryjliu: well it's still "cross-repo" in the sense that we're pip installing gpt-index which is not a part of this repo. if we had tests for every loader, this would catch issues where we remove stuff from BaseReader in gpt-index and it breaks loaders here. 

the main point i'm making is that we should mock download_loader to use the current files instead of doing a network request. happy to help with this tmrw

emptycrown:
Makes sense — I’ll take a look tomorrow

tests/test_file.py
jerryjliu: actually i thought about it, there's no point in using download_loader at all, since we'd be mocking it, and at that point it's easier to just import the class directly 

if we do want to test download_loader in a more e2e fashion, we could consider adding an integration test to the core gpt index repo (where download_loader lives). Then it will make a network request to llama-hub to download the corresponding loader; it would still test the logic of download_loader within gpt-index. Currently there's no integration tests in gpt-index, only unit tests, but this would be a good example of an initial integration test we should add..

emptycrown:
Sounds good. Yeah I was thinking that too last night. Will be a simple fix
here then

tests/test_file.py
```diff
@@ -0,0 +1,207 @@
+"""Test file reader."""
+import sys
+from tempfile import TemporaryDirectory
+from typing import Any, Dict
+
+sys.path.append("..")
```
jerryjliu, line 6: tbh i'm not a huge fan of having to add this to every test but also can't think of a better solution atm (an alternative would be to reintroduce a python package). so ok with this.

we could make this more explicit though:
```
sys.path.append(Path(__file__).parent.parent)
```

jerryjliu (approved):

tests/test_file.py
emptycrown: I considered reintroducing the package but it didn't seem like the right choice just for the sake of tests. Having the package without the intention of actually releasing LlamaHub as a package (which I think is not worth the effort to maintain rn) could introduce confusion and surface area for errors (e.g. people importing stuff incorrectly from within loaders).

I'll make that change though

PR #38, opened 2023-02-16
Add AsanaReader to load documents from an Asana workspace

daveey:


loader_hub/asana/base.py
```diff
@@ -0,0 +1,63 @@
...
+    """Asana reader. Reads data from an Asana workspace.
+
+    Args:
+        asana_token (str): Asana token.
+        asana_workspace (str): Asana workspace.
+    """
+
+    def __init__(self, asana_token: str) -> None:
+        """Initialize Asana reader."""
+        try:
```
emptycrown, line 19: No need to do a try/except here anymore since `requirements.txt` is installed by `download_loader` prior to this. You can just `import asana` here in the `__init__` function.

emptycrown (changes requested):
Looks good! Left a comment but the main thing is that you need to add AsanaReader to `library.json` in order for it to be available to `download_loader` and for it to appear on the website.

loader_hub/asana/base.py
daveey: Done

emptycrown (approved):

PR #43, opened 2023-02-16
Dad Jokes 🤓

SidU:
Because agents shouldn't just search on Google/Bing for what dad jokes are when someone wants a dad-joke. They should respond with actual dad-jokes!


emptycrown:
This is amazing haha. I'm going to use this for testing stuff in the future.

loader_hub/dadjokes/base.py
```diff
@@ -0,0 +1,32 @@
+"""icanhazdadjoke based dad-joke generator."""
```
emptycrown, line 1: Can we name this `dad_jokes` instead? Also, you'll need to add it to the `library.json` file for it to be picked up by `gpt_index.download_loader` or the website.

loader_hub/dadjokes/base.py
```diff
@@ -0,0 +1,32 @@
...
+
+    """
+
+    def _get_random_dad_joke(self):
+        response = requests.get("https://icanhazdadjoke.com/", headers={"Accept": "application/json"})
+        if response.status_code == 200:
+            json_data = response.json()
+            return json_data["joke"]
+        else:
+            return "Sorry, I couldn't get a joke right now. Please try again later."
```
emptycrown, line 22: Let's throw an error here instead of returning something that actually gets turned into a Document

emptycrown (changes requested):

SidU:
@emptycrown addressed your feedback 🙂

emptycrown (approved):

emptycrown:
Thanks -- I think you forgot to change the directory to `dad_jokes` as well?

SidU:
> Thanks -- I think you forgot to change the directory to `dad_jokes` as well?

Yes, just pushed a fix.

PR #46, opened 2023-02-17
Add tesseract model for plain text image

ravi03071991:
Hi,

Added tesseract model if the image has plain text and if the image has "key-value" pairs it uses DONUT model.

Thanks.

### DIFF
--- a/loader_hub/file/image/README.md
+++ b/loader_hub/file/image/README.md
@@ -1,6 +1,6 @@
 # Image Loader
 
-This loader extracts the text from an image that has text in it (e.g. a receipt). The [Donut](https://huggingface.co/docs/transformers/model_doc/donut) transformer model is used. The file extensions .png, .jpg, and .jpeg are preferred. A single local file is passed in each time you call `load_data`.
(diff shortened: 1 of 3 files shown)


loader_hub/file/image/base.py
```diff
@@ -19,78 +19,96 @@ class ImageReader(BaseReader):
...
-        except ImportError:
-            raise ValueError("sentencepiece is required for using DONUT model.")
-        try:
-            from PIL import Image  # noqa: F401
-        except ImportError:
-            raise ValueError(
-                "PIL is required to read image files." "Please run `pip install Pillow`"
+
+        if text_type == "plain_text":
+            try:
```
emptycrown, line 26: This isn't something you introduced, but we don't need to do try/except statements here anymore because requirements.txt is installed prior to this module. You should add `pytesseract` to the requirements.txt file and then remove the try/excepts.

loader_hub/file/image/README.md
```diff
@@ -12,7 +12,9 @@ from gpt_index import download_loader
 
 ImageReader = download_loader("ImageReader")
 
-loader = ImageReader()
+# If Image has key-value pairs text, use text_type = "key_value"
+# If Image has plain test, use use text_type = "plain_text" for plain text
```
emptycrown, line 16: Typo (test should be text)

emptycrown (changes requested):
Looks nice! Added a couple comments.

loader_hub/file/image/base.py
ravi03071991: Got it. Updated.

emptycrown (approved):

PR #49, opened 2023-02-20
swap out gpt_index imports for llama_index imports

jerryjliu:
I'm thinking we could land this tomorrow or Tuesday - the main implication here is that users will need to `pip install llama_index` in order to use llama-hub, since all loaders now have llama_index deps.


emptycrown:
Sounds good. There are a few mentions of "GPT Index" in READMEs, etc. I feel like we should do a replace-all for them too?

emptycrown (approved):

jerryjliu:
let's merge this later tonight if that's ok, going to line up a release

PR #53, opened 2023-02-21
Add Memos loader

p3psi-boo:
For [Memos](https://github.com/usememos/memos).


loader_hub/memos/base.py
```diff
@@ -0,0 +1,54 @@
...
+    """Memos reader.
+
+    Reads content from an Memos.
+
+    """
+
+    def __init__(self, host: str = "https://demo.usememos.com/") -> None:
+        """Init params."""
+        
+        self._memoUrl = host + "api/memo"
```
emptycrown, line 19: Can we use `urljoin` here? It accounts for extra/missing slashes: https://docs.python.org/3/library/urllib.parse.html

loader_hub/memos/base.py
```diff
@@ -0,0 +1,54 @@
...
+
+        """
+        import requests
+        
+        documents = []
+
+        realUrl = self._memoUrl
+        if ~params:
+            params = {}
+            realUrl += "/all"
```
emptycrown, line 38: Same here: `urljoin` is best.

loader_hub/memos/base.py
```diff
@@ -0,0 +1,54 @@
...
+
+        req = requests.get(realUrl, params)
+        
+        if ~req.ok:
```
emptycrown, line 42: It's probably better to raise an error here instead of returning an empty list

emptycrown (changes requested):
Thanks for contributing this! Left a couple of comments.

p3psi-boo:
Fixed it.
Thanks for your suggestions.

p3psi-boo:
Wait a min

loader_hub/memos/base.py
```diff
@@ -31,21 +32,16 @@ def load_data(self, params: Optional[Any] = 101) -> List[Document]:
...
         realUrl = self._memoUrl
+        
         if ~params:
             params = {}
-            realUrl += "/all"
+            realUrl = urljoin(self._memoUrl, "/all")
 
         req = requests.get(realUrl, params)
-        
-        if ~req.ok:
```
emptycrown: Err I meant if request is not ok, `raise ValueError("Your Memo URL is not valid")` or something like that

emptycrown (changes requested):

loader_hub/memos/base.py
p3psi-boo: Ok

emptycrown (approved):

emptycrown (approved):

PR #58, opened 2023-02-23
Add Google Calendar Loader

ong:
Add a Google Calendar loader, using roughly the same format as @jerryjliu's Google Docs reader.

It works pretty well for free form Q&A for one's schedule!

 Screenshot 2023-02-22 at 11 41 16 PM


emptycrown (commented):
This looks solid! Thanks for putting this together.

One comment: does it make sense to include past events as well? Or at least have a setting for that.

ong:
> One comment: does it make sense to include past events as well? Or at least have a setting for that.

Updated to include an optional `start_date` argument to retrieve past events.

emptycrown (approved):

PR #59, opened 2023-02-23
Add Spotify Loader

ong:
Add a Spotify loader that reads data from your account. Might be useful for music discovery/recommendation use cases.

Screenshot 2023-02-23 at 2 57 29 PM


emptycrown (approved):
Thanks! Cool idea.

PR #62, opened 2023-02-25
Add RemoteDepth to llamahub

simonManydata:
Needed to build a reader where we download all the links available in a webpage (also links inside links).

Difficult to write tests without network requests. 

Not sure if I should have extended the reader `RemoteReader`, modify existing  `RemoteReader`, or just create a new class from BaseClass.

### DIFF
--- a/loader_hub/remote_depth/base.py
+++ b/loader_hub/remote_depth/base.py
@@ -1,7 +1,6 @@
 """Remote file reader.
 
-A loader that fetches an arbitrary remote page or file by URL and parses its contents.
-
+A loader that fetches any remote page or file by URL and retrieves child pages with certain constraints. The class also parses the contents of each page and provides access to the parsed data.
 """
 from typing import Any, Dict, List, Optional, Union
 
--- a/loader_hub/remote_depth/base.py
+++ b/loader_hub/remote_depth/base.py
@@ -4,17 +4,15 @@ A loader that fetches an arbitrary remote page or file by URL and parses its con
 
 """
 from typing import Any, Dict, List, Optional, Union
-from urllib.parse import urljoin, urlparse, urlunparse
 
 import requests
-from bs4 import BeautifulSoup
 from llama_index import download_loader
 from llama_index.readers.base import BaseReader
 from llama_index.readers.schema.base import Document
-from tqdm.auto import tqdm
 
 
 class RemoteDepthReader(BaseReader):
+
     def __init__(
         self,
         *args: Any,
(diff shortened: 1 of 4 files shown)


loader_hub/remote_depth/base.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+from typing import Any, Dict, List, Optional, Union
+from urllib.parse import urljoin, urlparse, urlunparse
+
+import requests
+from bs4 import BeautifulSoup
+from llama_index import download_loader
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+from tqdm.auto import tqdm
```
emptycrown, line 14: Can you lazily import these packages from within the class? For now, it makes it easier to test and also download many loaders at once. Additionally, please create a `requirements.txt` in this directory and drop in bs4 and tqdm.

loader_hub/remote_depth/base.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class RemoteDepthReader(BaseReader):
```
emptycrown, line 17: I think it's ok to have this as a separate class for now. IMO it's closer to a web scraper than `RemoteReader`.

loader_hub/remote_depth/base.py
```diff
@@ -0,0 +1,90 @@
+"""Remote file reader.
+
+A loader that fetches an arbitrary remote page or file by URL and parses its contents.
```
emptycrown, line 3: Update this description.

emptycrown (changes requested):
Thanks! Left a few comments but looks good overall.

loader_hub/remote_depth/base.py
simonManydata: Yes, correct, Should we rename that to `WebScraperReader` when we have more functionalities, and keep that named `RemoteDepthReader` for now?

loader_hub/remote_depth/base.py
simonManydata: I have lazily imported the packages `bs4, `tqdm` and `urllib.parse`

emptycrown (approved):

loader_hub/remote_depth/base.py
emptycrown: Yeah let's keep it for now. There are other web scrapers in the library already. We can reorganize down the road if needed.

simonManydata:
Still on the 24 from PST!

emptycrown:
Yep you're in the pool :)

PR #63, opened 2023-02-25
Add RDF file loader

mommi84:
Adding [RDF](https://en.wikipedia.org/wiki/Resource_Description_Framework) file loader to convert standardised graph data into text-based triples.

Example:
```
<Milan> <region> <Lombardy>
<city> <subClassOf> <place>
<Lombardy> <type> <region>
<region> <subClassOf> <place>
<Milan> <type> <city>
<place> <subClassOf> <thing>
<Milan> <mayor> <Beppe Sala>
<Beppe Sala> <type> <person>
```


loader_hub/file/rdf/base.py
```diff
@@ -0,0 +1,65 @@
+"""Read RDF files."""
+
+from pathlib import Path
+from typing import Dict, List, Optional
+
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+
+from rdflib import Graph, URIRef
```
emptycrown, line 9: Can you lazily import these from within the class? Makes testing easier and makes things cleaner when we download a bunch of loaders at once.

emptycrown (changes requested):
Thanks for this! Looks great -- just added one comment.

emptycrown (approved):

loader_hub/file/rdf/base.py
mommi84: I just saw you already did it. Thanks!

PR #64, opened 2023-02-26
Add more patterns url for YoutubeTranscriptReader

simonManydata:
Sometimes when I am scraping the web I encounter `/embed`, so just creating a new pattern to extract youtube url like this.


emptycrown (commented):
Cool, nice addition! Is it possible for Youtube links to not have the full `https://www.youtube.com` at the start? Like what if it's just `https://youtube.com`?

emptycrown (commented):
See above -- I'm not sure if the regex patterns are complete.

emptycrown (approved):

PR #65, opened 2023-02-26
adding youtube url in the base remote reader

simonManydata:
Improving the RemoteReader, let's read YT videos!


emptycrown (approved):
Cool! Good way to reuse loaders together. We can probably rethink a more extendable way to do this down the line so that it can apply to more than using Youtube links but this is a good start!

PR #68, opened 2023-02-26
Fix Notion Loader: For a Database we're only loading the first 100 page_ids

MarkusOdenthal:
During testing of the Notion Loader today, I noticed that I only received the first 100 pages when loading all pages from a database. Notion API returns the attribute has_more, and if this is true, we must request more pages. So this is the implementation of the fix.

### DIFF
--- a/loader_hub/notion/base.py
+++ b/loader_hub/notion/base.py
@@ -90,21 +90,32 @@ class NotionPageReader(BaseReader):
         """Read a page."""
         return self._read_block(page_id)
 
-    def query_database(
-        self, database_id: str, query_dict: Dict[str, Any] = {}
-    ) -> List[str]:
+    def query_database(self, database_id: str) -> List[str]:
         """Get all the pages from a Notion database."""
+        query_dict = {"page_size": 100}
+        pages = []
+
         res = requests.post(
             DATABASE_URL_TMPL.format(database_id=database_id),
             headers=self.headers,
             json=query_dict,
         )
-        data = res.json()
-        page_ids = []
-        for result in data["results"]:
-            page_id = result["id"]
-            page_ids.append(page_id)
+        if res.ok:
+            data = res.json()
+            pages.extend(data.get("results"))
 
+            while data.get("has_more"):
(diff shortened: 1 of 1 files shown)


loader_hub/notion/base.py
```diff
@@ -90,21 +90,32 @@ def read_page(self, page_id: str) -> str:
...
             DATABASE_URL_TMPL.format(database_id=database_id),
             headers=self.headers,
             json=query_dict,
         )
-        data = res.json()
-        page_ids = []
-        for result in data["results"]:
-            page_id = result["id"]
-            page_ids.append(page_id)
+        if res.ok:
```
emptycrown, line 103: I think it's better to just raise an error if the result is not ok instead of just returning an empty list.

loader_hub/notion/base.py
```diff
@@ -90,21 +90,32 @@ def read_page(self, page_id: str) -> str:
         """Read a page."""
         return self._read_block(page_id)
 
-    def query_database(
-        self, database_id: str, query_dict: Dict[str, Any] = {}
-    ) -> List[str]:
+    def query_database(self, database_id: str) -> List[str]:
         """Get all the pages from a Notion database."""
+        query_dict = {"page_size": 100}
```
emptycrown, line 95: I think this approach is fine but it's best to keep query_dict as an argument you can pass in still. Perhaps it can just default to {"page_size": 100}

emptycrown (changes requested):
Thanks for doing this! Left a couple of comments.

loader_hub/notion/base.py
MarkusOdenthal: Good point thanks.

I implemented it using the raise_for_status() method

Eventually, we can add error handling: https://developers.notion.com/reference/errors

In this way, the user will know better what kind of error is being raised. If you'd like, I can do it next week.

loader_hub/notion/base.py
MarkusOdenthal: I agree :)

emptycrown (approved):

PR #71, opened 2023-02-28
Update for MongoDB Atlast

Rishav-hub:
Previously class took `host` and `port` as a parameter. Connect with MongoDB Atlas it's not possible to provide the host and port as it has a single URL. So directly pass the URL to the MongoClient.

### DIFF
--- a/loader_hub/mongo/base.py
+++ b/loader_hub/mongo/base.py
@@ -17,7 +17,7 @@ class SimpleMongoReader(BaseReader):
 
     """
 
-    def __init__(self, host: str, port: int, mongo_db_url: Optional[Dict] = None  max_docs: int = 1000) -> None:
+    def __init__(self, host: str, port: int, mongo_db_url: Optional[Dict] = None, max_docs: int = 1000) -> None:
         """Initialize with parameters."""
         from pymongo import MongoClient  # noqa: F401
 
--- a/loader_hub/mongo/base.py
+++ b/loader_hub/mongo/base.py
@@ -17,11 +17,14 @@ class SimpleMongoReader(BaseReader):
 
     """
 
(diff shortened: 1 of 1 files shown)


Rishav-hub:
@emptycrown

loader_hub/mongo/base.py
```diff
@@ -12,17 +12,16 @@ class SimpleMongoReader(BaseReader):
     Concatenates each Mongo doc into Document used by LlamaIndex.
 
     Args:
-        host (str): Mongo host.
-        port (int): Mongo port.
+        mongo_db_url (str): Mongo Full URL.
```
emptycrown, line 15: how about just db_url or url?

emptycrown (changes requested):
Good catch. How about we make it so both options are usable? So `url` is a new `Optional[str]` and it uses that arg if host/port isn't available.

Rishav-hub (commented):
Yes that would also work.

jerryjliu:
hi @Rishav-hub, are you able to get to this? if not i can help land

loader_hub/mongo/base.py
Rishav-hub: db_url would work

Rishav-hub:
> hi @Rishav-hub, are you able to get to this? if not i can help land

Actually little confused where can I again change the code in this pull request.

jerryjliu:
> > hi @Rishav-hub, are you able to get to this? if not i can help land
> 
> Actually little confused where can I again change the code in this pull request.

as in, keep the host/port parameters, and set mongo_db_url as an Optional string. if mongo_db_url is specified use that, otherwise use host and port

Rishav-hub (commented):
added condition for mongodb url.

jerryjliu (approved):

emptycrown (approved):

PR #73, opened 2023-03-01
GithubRepositoryReader tests and minor fixes

ahmetkca:
Added test for `GithubRepositoryReader` and other parts of the code that `GithubRepositoryReader` uses additional to the `GithubClient`.

Also, there was a minor fix for directory file extension filter where it wasn't filtering properly.

I also tested in an isolated environment and I wasn't able to download the `GithubRepositoryReader` but it should be fixed by this PR https://github.com/jerryjliu/gpt_index/pull/574


jerryjliu:
@ahmetkca thanks for PR, will review soon!

loader_hub/github_repo/base.py
```diff
@@ -5,25 +5,50 @@
...
 from llama_index.readers.file.base import DEFAULT_FILE_EXTRACTOR
-from llama_index.readers.llamahub_modules.github_repo.github_client import (
-    BaseGithubClient, GitBranchResponseModel, GitCommitResponseModel,
-    GithubClient, GitTreeResponseModel)
-from llama_index.readers.llamahub_modules.github_repo.utils import (
-    BufferedGitBlobDataIterator, get_file_extension, print_if_verbose)
 from llama_index.readers.schema.base import Document
 
+
+if "pytest" in sys.modules:
```
jerryjliu, line 24: ah i see. this is to get unit tests to pass?

loader_hub/github_repo/base.py
```diff
@@ -123,30 +148,33 @@ def _check_filter_directories(self, tree_obj_path: str) -> bool:
...
-                tree_obj_path.startswith(directory)
-                or directory.startswith(tree_obj_path)
-                for directory in filter_directories
-            )
-        else:
-            raise ValueError(
-                f"Unknown filter type: {filter_type}. "
-                "Please use either 'ignore' or 'include'."
-            )
+        match filter_type:
```
jerryjliu, line 160: this would break backwards compatibility with older python versions (only supported on python 3.10), let's keep as if/else instead

jerryjliu (commented):
sorry @ahmetkca for the long delay, thanks for putting out this PR! one comment around still supporting older python versions. also does the unit test pass for you?

loader_hub/github_repo/base.py
ahmetkca: yeah, Unfortunately I couldn't find a nicer way to do this

loader_hub/github_repo/base.py
ahmetkca: > sorry @ahmetkca for the long delay, thanks for putting out this PR! one comment around still supporting older python versions. also does the unit test pass for you?

It does but for GitHub client test you need to set `GITHUB_TOKEN` env variable which I honestly don't like but it makes network requests to check the test case.

ahmetkca: > this would break backwards compatibility with older python versions (only supported on python 3.10), let's keep as if/else instead
btw What is the oldest python version LlamaIndex supports? so I can test it on that version which should cover higher versions.

ahmetkca:
@jerryjliu ca we revert this. The filter logic was not working properly. I think I fixed and added a better test for it.

jerryjliu:
> @jerryjliu ca we revert this. The filter logic was not working properly. I think I fixed and added a better test for it.

sure! np

jerryjliu:
@ahmetkca  https://github.com/emptycrown/llama-hub/pull/100

let me know when you have the updated PR out, can add that in

PR #79, opened 2023-03-02
Added reference links to readmedocs reader

Smyja:
I have added reference links to the reader and resolved a value error issue i didn't notice.

### DIFF
--- a/loader_hub/web/beautiful_soup_web/base.py
+++ b/loader_hub/web/beautiful_soup_web/base.py
@@ -62,14 +62,24 @@ def _readmedocs_reader(soup: Any, url: str) -> Tuple[str, Dict[str, Any]]:
         page_link = requests.get(doc_link)
         soup = BeautifulSoup(page_link.text, "html.parser")
         try:
-            text = soup.find_all("article", {"id": "content"})[0].get_text()
+            text = ""
+            for element in soup.find_all("article", {"id": "content"}):
+                for child in element.descendants:
+                    if child.name == "a" and child.has_attr("href"):
+                        url = child.get('href')
+                        if url is not None and "edit" in url:
+                            text += child.text
+                        else:
+                            text += f"{child.text} (Reference url: {doc_link}{url}) "
+                    elif child.string:
+                        text += child.string.strip() + " "
         except IndexError:
             text = None
-            raise IndexError("No content found")
-
-        if text:
-            texts.append("\n".join([t for t in text.split("\n") if t]))
+            print(f"Could not find article with id 'content' in {doc_link}")
+            continue
+        texts.append("\n".join([t for t in text.split("\n") if t]))
     return "\n".join(texts), {}
+
 
 
 DEFAULT_WEBSITE_EXTRACTOR: Dict[


loader_hub/web/beautiful_soup_web/base.py
```diff
@@ -62,16 +62,26 @@ def _readmedocs_reader(soup: Any, url: str) -> Tuple[str, Dict[str, Any]]:
...
+                            text += f"{child.text} (Reference url: {doc_link}{url}) "
+                    elif child.string:
+                        text += child.string.strip() + " "
         except IndexError:
             text = None
-            raise IndexError("No content found")
-
-        if text:
-            texts.append("\n".join([t for t in text.split("\n") if t]))
+            print(f"Could not find article with id 'content' in {doc_link}")
```
jerryjliu, line 78: let's use logger and not print (to be consistent)

jerryjliu (commented):
thanks for submitting this! I'm not sure what you end use case is, but besides adding the link to the raw text itself, wouldn't you also want to add the link as a metadata field that gets injected on the Document object?

loader_hub/web/beautiful_soup_web/base.py
Smyja: oh, i'll do that.

Smyja:
> thanks for submitting this! I'm not sure what you end use case is, but besides adding the link to the raw text itself, wouldn't you also want to add the link as a metadata field that gets injected on the Document object?

This allows you to get reference links for all readme documentations like kapa.ai. Just request for the link in your prompt and it will provide it. So this doesn't need to be in the metadata field since the llm can't tell which page or text owns which link.

loader_hub/web/beautiful_soup_web/base.py
Smyja: I've added it.

loader_hub/web/beautiful_soup_web/base.py
```diff
@@ -62,13 +66,22 @@ def _readmedocs_reader(soup: Any, url: str) -> Tuple[str, Dict[str, Any]]:
         page_link = requests.get(doc_link)
         soup = BeautifulSoup(page_link.text, "html.parser")
         try:
-            text = soup.find_all("article", {"id": "content"})[0].get_text()
+            text = ""
+            for element in soup.find_all("article", {"id": "content"}):
+                for child in element.descendants:
+                    if child.name == "a" and child.has_attr("href"):
```
jerryjliu, line 72: My one last request is - can we make this depend on a config parameter `include_url_in_text` in the __init__.py function that could be set to True by default? gives us extensability in case users want to include these links for other domain-specific parsers

Smyja: Okay, I'll look into to it.

Smyja: I've updated it.

jerryjliu: thanks. yeah i meant having it as an option in the `__init__` function (and then passing those params to these functions like _readmedocs_reader)

jerryjliu (approved):
thanks @Smyja. no worries i can also fix this later

PR #87, opened 2023-03-06
Relax fixed requirements

oraluben:
Some requirements are fixed to a specific version and will fail if even minor versions do not match.
This PR resolves those dependencies a little bit. (But could be even wider if possible)

Example error before fix:
```
$ python main.py                 
Traceback (most recent call last):
  File "/Users/yyc/PycharmProjects/idx/main.py", line 3, in <module>
    PythonDocReader = download_loader('BeautifulSoupWebReader')
  File "/Users/yyc/.local/miniforge3/lib/python3.10/site-packages/llama_index/readers/download.py", line 190, in download_loader
    pkg_resources.require([str(r) for r in requirements])
  File "/Users/yyc/.local/miniforge3/lib/python3.10/site-packages/pkg_resources/__init__.py", line 909, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/yyc/.local/miniforge3/lib/python3.10/site-packages/pkg_resources/__init__.py", line 800, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (urllib3 1.26.14 (/Users/yyc/.local/miniforge3/lib/python3.10/site-packages), Requirement.parse('urllib3==1.26.13'))
```


loader_hub/reddit/requirements.txt
```diff
@@ -1,9 +1,5 @@
-certifi==2022.12.7
```
jerryjliu, line 1: @oraluben why remove these requirements?

jerryjliu (commented):
thanks! sorry for the delay. happy to merge this soon

loader_hub/reddit/requirements.txt
oraluben: This will be installed via requests and I see no need to keep it.

jerryjliu: ah ok, makes sense

jerryjliu (approved):

PR #88, opened 2023-03-06
Added excel loader

maccarini:
Pandas parser for .xlsx files.


jerryjliu:
> Pandas parser for .xlsx files.

sweet! will take a look soon

jerryjliu (commented):
looks good! only remaining thing is, could you add an entry to library.json?

PR #90, opened 2023-03-07
Update base.py

SergeWilfried:
Fix: PDF reader


loader_hub/file/pdf/base.py
```diff
@@ -19,7 +19,7 @@ def load_data(
         text_list = []
         with open(file, "rb") as fp:
             # Create a PDF object
-            pdf = PyPDF2.PdfReader(fp)
+            pdf = PyPDF2.PdfFileReader(fp)
```
jerryjliu, line 22: isn't PdfFileReader deprecated?

PR #91, opened 2023-03-07
Fix to support the latest version of langchain

masatootake:
I've tried to fix https://github.com/emptycrown/llama-hub/issues/83


jerryjliu (approved):
thanks!

PR #94, opened 2023-03-10
feat: add readability webpage loader

pandazki:
Extracting relevant information from a fully rendered web page.
During the processing, it is always assumed that web pages used as data sources contain textual content.

It is particularly effective for websites that use client-side rendering.

1. Load the page and wait for it fully loaded. (playwright)
2. Inject Readability.js to extract the main content.


pandazki:
Why not use the Knowledge Base Website Loader? 
I wanted to quickly summarize a webpage's content, so I read the documentation and tried using it. However, I found that providing the article path and various selectors was extremely inconvenient. A purely automated version would be more suitable for me.

jerryjliu:
nice! so to summarize, to make sure i understand this correctly:
- automated web page scraping
- what does the "fully rendered" web page have to do with this? don't all scrapers operate over fully rendered pages?

pandazki:
> nice! so to summarize, to make sure i understand this correctly:
> 
> * automated web page scraping
> * what does the "fully rendered" web page have to do with this? don't all scrapers operate over fully rendered pages?

Yes, automated web page scraping is the primary objective. 
Regarding the second point, I have encountered different web pages that cannot be fully rendered using the same settings. Therefore, I have provided [the settings](https://github.com/pandazki/llama-hub/blob/2cf33990ee6ae5005e3cbbd55ea52093626392bd/loader_hub/web/fully_rendered_web/README.md?plain=1#L21) to customize the waiting strategy. 
I admit that 'fully rendered webpage loader' is not a good name😳. Do you have any suggestions?

jerryjliu:
> Yes, automated web page scraping is the primary objective. Regarding the second point, I have encountered different web pages that cannot be fully rendered using the same settings. Therefore, I have provided [the settings](https://github.com/pandazki/llama-hub/blob/2cf33990ee6ae5005e3cbbd55ea52093626392bd/loader_hub/web/fully_rendered_web/README.md?plain=1#L21) to customize the waiting strategy. I admit that 'fully rendered webpage loader' is not a good name😳. Do you have any suggestions?

It could be called ReadabilityJSReader? Since the main goal of this is to load a web page and use ReadabilityJS to scrape the main content of the page

pandazki:
> > Yes, automated web page scraping is the primary objective. Regarding the second point, I have encountered different web pages that cannot be fully rendered using the same settings. Therefore, I have provided [the settings](https://github.com/pandazki/llama-hub/blob/2cf33990ee6ae5005e3cbbd55ea52093626392bd/loader_hub/web/fully_rendered_web/README.md?plain=1#L21) to customize the waiting strategy. I admit that 'fully rendered webpage loader' is not a good name😳. Do you have any suggestions?
> 
> It could be called ReadabilityJSReader? Since the main goal of this is to load a web page and use ReadabilityJS to scrape the main content of the page

Nice idea! 
I have renamed it as ReadabilityWebPageReader.

PR #97, opened 2023-03-10
isolating abstracts from full papers

reletreby:
- The abstract is already included in the main paper, so returning papers Documents only won't lead to loss of abstract information
- User could be willing to only look into abstracts, hence it is convenient to isolate abstracts from full papers


jerryjliu (changes requested):
i like the idea! main concern is breaking backwards compatibility.

is it possible to define another function `load_papers_and_abstracts` that contains the main logic (returns both arxiv docs and abstract docs), but still keep the `load_data` definition that returns them joined together?

reletreby:
You got it @jerryjliu !

jerryjliu (approved):

PR #108, opened 2023-03-13
Improvements to Gmail Reader

reletreby:
- Fix random port assignment (similar to this [PR](https://github.com/emptycrown/llama-hub/pull/96))
- Build an iterative function that handles MIME `multipart` objects
      - some emails consist of `multipart` objects such as text and attached PDF, etc.
      - More on MIME `multipart` can be found [here](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html)
      - I basically iterate over all objects recursively until I catch plain text, and when I do, I retrieve that

I have this method disabled by default to leave the current approach as is, but I actually recommend we make this iterative method the default one. I have tested the logic and all looks good on my end _(i.e., I can receive clean plain text)_.

### DIFF
--- a/loader_hub/gmail/base.py
+++ b/loader_hub/gmail/base.py
@@ -19,6 +19,7 @@ class GmailReader(BaseReader, BaseModel):
         max_results (int): Max number of results. Defaults to 10.
     """
     query: str = None
+    use_iterative_parser: bool = False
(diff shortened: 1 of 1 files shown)


jerryjliu (approved):
amazing!

PR #113, opened 2023-03-15
Added Zendesk Ticket reader

robert4dt:
I will also add a crawler for comments later


jerryjliu:
@robert4dt out of curiosity why close?

robert4dt:
Hi,
I will re merge again. I found some issues: OpenAI won't be able to handle such a huge amount of data
________________________________

PR #115, opened 2023-03-15
Added domain lock paramater

cdstrachan:
Existing code will follow any link it finds, regardless of where it goes. Now if domain_lock is TRUE, it will only index pages on the provided domain. EG if your domain is https://www.test.com, and it finds a link to https://www.test2.com, it will be discarded of domain_lock is true. Its optional and defaults to FALSE, so non-breaking change.

### DIFF
--- a/loader_hub/remote_depth/base.py
+++ b/loader_hub/remote_depth/base.py
@@ -39,9 +39,7 @@ class RemoteDepthReader(BaseReader):
             urls[i] = []
             new_links = []
             print(f"Reading links at depth {i}...")
-
             for link in tqdm(links):
-                print(link)
                 """Checking if the link belongs the provided domain. """
                 if ((self.domain_lock and link>url) or (not self.domain_lock)):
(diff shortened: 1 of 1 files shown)


loader_hub/remote_depth/base.py
```diff
@@ -39,12 +40,17 @@ def load_data(self, url: str) -> List[Document]:
...
             for link in tqdm(links):
-                if link in links_visited:
-                    continue
-                if link:
-                    urls[i].append(link)
-                    new_links.extend(self.get_links(link))
-                links_visited.append(link)
+                """Checking if the link belongs the provided domain. """
+                if ((self.domain_lock and link.find(url)>-1) or (not self.domain_lock)):
+                    print("Loading link: " + link)
```
jerryjliu, line 45: ubernit: remove print (let's try to use logger instead)

jerryjliu (approved):

PR #123, opened 2023-03-16
Iterative Parser for Gmail

reletreby:
Enabling iterative parser as default option for Gmail reader. I have tested it many times and I think it works better than current default method.


jerryjliu (commented):
@reletreby does this make loading times slower?

logan-markewich:
Going to close this PR, users cam just set the option as needed.

PR #137, opened 2023-03-24
feat: Add OpendalReader and add s3|azblob|gcs support

Xuanwo:
Refer to https://github.com/jerryjliu/llama_index/issues/865 for more details.

This PR only adds s3|azblob|gcs support and I'm planning to add oss|obs|webhdfs|... services.

## Existing Problem

Many file extractors only accept local paths, which means we have to download the file to a temporary directory first. This can be a huge burden if our dataset is large. To address this issue, we should support accepting a file-like object so that we can read data directly from remote storage services.

## Future possibilities

With OpenDAL, we can add the following services support while needed:

- azblob: Azure Storage Blob services.
- azdfs: Azure Data Lake Storage Gen2 services. (As known as abfs)
- fs: POSIX alike file system.
- ftp: FTP and FTPS support.
- gcs: Google Cloud Storage Service.
- hdfs: Hadoop Distributed File System(HDFS).
- http: HTTP read-only services.
- ipfs: InterPlanetary File System HTTP Gateway support.
- ipmfs: InterPlanetary File System MFS API support.
- memcached: Memcached service support.
- obs: Huawei Cloud Object Storage Service (OBS).
- oss: Aliyun Object Storage Service (OSS).
- redis: Redis services support.
- s3: AWS S3 alike services.
- webdav: WebDAV Service Support.
- webhdfs: WebHDFS Service Support.

## Test with local minio

```py
import logging
logging.basicConfig(level=logging.DEBUG)

import os
os.environ["OPENAI_API_KEY"] = 'YOUR_OPENAI_API_KEY'

from llama_index import GPTSimpleVectorIndex, SimpleDirectoryReader, download_loader


OpendalReader = download_loader(
    'OpendalReader',
    loader_hub_url="https://raw.githubusercontent.com/Xuanwo/llama-hub/a43d73c1745c5e4285e5ccf0daad7b92b330b22a/loader_hub",
    refresh_cache=True
)

loader = OpendalReader(
    scheme='s3',
    path = '/',
    bucket='opendal',
    endpoint='http://127.0.0.1:9900',
    access_key_id='minioadmin',
    secret_access_key='minioadmin',
)
documents = loader.load_data()
```


jerryjliu (approved):
this is awesome!!

loader_hub/opendal/base.py
```diff
@@ -0,0 +1,79 @@
...
+"""
+import tempfile
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Union
+import asyncio
+
+from llama_index import download_loader
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+import opendal
```
jerryjliu, line 14: actually these imports should probably be lazily loaded so that it doesn't break loading for other modules (otherwise unit tests also currently fail)

Xuanwo:
Thanks for the change (and the quick merge)!

jerryjliu:
> Thanks for the change (and the quick merge)!

thank you for implementing!

PR #138, opened 2023-03-24
feat: ReadabilityWebPageReader add normalize func and custom TextSplitter

pandazki:
- normalize function is used to normalize whitespace characters, and nfkc is the default option.
- Text splitter is used to automatically split long texts.

### DIFF
--- a/loader_hub/web/readability_web/base.py
+++ b/loader_hub/web/readability_web/base.py
@@ -1,8 +1,10 @@
-from typing import Dict, List, Literal, Optional, Any, cast
+from typing import Callable, Dict, List, Literal, Optional, Any, cast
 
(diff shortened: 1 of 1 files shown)


loader_hub/web/readability_web/base.py
```diff
@@ -77,9 +87,16 @@ def load_data(self, url: str) -> List[Document]:
                 "siteName",
             ]}
 
+            article["textContent"] = self._normalize(article["textContent"])
```
jerryjliu, line 90: if self._normalize allowed to be None? if so we should add a check to see if it's none first

pandazki:
Thx for review~ All fixed.

jerryjliu (approved):

PR #142, opened 2023-03-27
feat: Added an option to choose the language in Wikipedia texts

emmanuel-oliveira:
As GPT is multilingual, I think it's important to be able to choose the language in the download loader whenever possible, so I've added an option to choose language from Wikipedia texts


jerryjliu (approved):
nice!

PR #148, opened 2023-03-29
feat: Created a loader for flattened pdfs

emmanuel-oliveira:
The idea of this loader is for flat pdfs or pdfs that are just images, the existing loader for pdf cannot identify the texts in these types of pdf, I thought it would be better to be a new loader than to edit the existing one because it is a slower process to identify the texts.

The logic is to read a pdf, convert all its pages into images and use the already existing image loader in llama-hub to extract the text.


jerryjliu:
this is a cool idea! will take a deeper look soon

loader_hub/file/flat_pdf/base.py
```diff
@@ -0,0 +1,80 @@
...
+import warnings
+from pathlib import Path
+
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+
+
+class FlatPdfReader(BaseReader):
+
+    pdf_dir: Path
```
jerryjliu, line 13: won't this error since it's a class attribute? it's not shown in your README how to specify these args (`pdf_dir`, `pdf_name`, `pdf_pages_count`, etc.)

jerryjliu (commented):
generally looks pretty good, one main question about the class attributes

loader_hub/file/flat_pdf/base.py
emmanuel-oliveira: They were only being used internally, a "global variable" logic, but with your comment, I looked again and with that logic it may cause strange behavior as the `FlatPdfReader` object can be instantiated once and used for many different pdfs , so I made the correction for these attributes to become local scope variables in the methods.

jerryjliu (approved):

PR #159, opened 2023-04-01
Add Feedly Rss Reader

kychanbp:
https://user-images.githubusercontent.com/17669769/229293705-df9b0743-3af4-4bc6-a211-31113c00ce9c.mov

The loader uses wrapper created by [Feedly](https://github.com/feedly/python-api-client/blob/master/examples/stream_entries.py) to get RSS Feeds subscribed in Feedly.


jerryjliu (approved):
thanks!

loader_hub/feedly_rss/base.py
```diff
@@ -0,0 +1,56 @@
+"""Feedly Rss Reader"""
+
+from pathlib import Path
+from feedly.api_client.session import FeedlySession
```
jerryjliu, line 4: oh since it's failing tests, the one comment i was going to make is to move this to be a lazy import (e.g. in `load_data`)

kychanbp:
@jerryjliu Do you think it is better to construct the `Document` as `string` instead of `dict`? I tried to construct index of the documents and give me `AttributeError: 'dict' object has no attribute 'split'`?

PR #160, opened 2023-04-02
Rename JSONReader.load_data input_file param to file to match other file reader load_data signatures

saqadri:
Currently, when JSONReader gets invoked by `SimpleDirectoryReader`, it results in the following error:
```
JSONReader.load_data() got an unexpected keyword argument 'file'
```

This is due to the `file` param not being present in `JSONReader.load_data`. This fix makes the reader's `load_data` function signature consistent with the other file readers.


jerryjliu (approved):
nice catch!! thanks 🙏

PR #164, opened 2023-04-05
Add transcription using gladia

ravi03071991:
Hi,

Added audio loader using Gladia which uses whisper backend with optimizations. It usually takes less than a minute to get the transcription for an hour audio file.

Thank you.

### DIFF
--- a/loader_hub/file/audio/base.py
+++ b/loader_hub/file/audio/base.py
@@ -17,40 +17,23 @@ class AudioTranscriber(BaseReader):
 
     """
 
-    def __init__(self, *args: Any, model_version: str = "base", 
-                model_type: str = "openai_whisper", diarization_max_speakers: str = None,
-                language: str = None, language_behaviour: str = "automatic multiple languages",
-                target_translation_language: str = 'english', gladia_api_key: str = None, 
-                transcription_hint: str = None, **kwargs: Any) -> None:
+    def __init__(self, *args: Any, model_version: str = "base", **kwargs: Any) -> None:
         """Init params."""
         super().__init__(*args, **kwargs)
+        self._model_version = model_version
 
-        self.parser_config = {}
-        self.model_type = model_type
+        import whisper
 
-        if self.model_type == "gladia":
-            self.parser_config['gladia_api_key'] = gladia_api_key
-            self.parser_config['diarization_max_speakers'] = diarization_max_speakers
-            self.parser_config['language'] = language
-            self.parser_config['language_behaviour'] = language_behaviour
-            self.parser_config['target_translation_language'] = target_translation_language
-            self.parser_config['transcription_hint'] = transcription_hint
-        elif self.model_type == "openai_whisper":
-            self._model_version = model_version
+        model = whisper.load_model(self._model_version)
 
-            import whisper
-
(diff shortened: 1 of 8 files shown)


loader_hub/file/audio/base.py
```diff
@@ -17,16 +17,32 @@ class AudioTranscriber(BaseReader):
...
+        self.model_type = model_type
+
+        if self.model_type == "openai_gladia_whsiper":
+            self.parser_config['gladia_api_key'] = gladia_api_key
+            self.parser_config['diarization_max_speakers'] = diarization_max_speakers
+            self.parser_config['language'] = language
+            self.parser_config['language_behaviour'] = language_behaviour
+            self.parser_config['target_translation_language'] = target_translation_language
+            self.parser_config['transcription_hint'] = transcription_hint
+        else:
```
jerryjliu, line 38: let's do `elif` model_version == "gladia", and then explicitly raise ValueError in the `else` block if the model_version is invalid

loader_hub/file/audio/base.py
```diff
@@ -43,13 +59,41 @@ def load_data(
...
+            audio.export(file, format="mp3")
+
+        transcript = ""
+
+        if self.model_type == "openai_whisper":
+            model = cast(whisper.Whisper, self.parser_config["model"])
+            result = model.transcribe(str(file))
+
+            transcript = result["text"]
+        else:
```
jerryjliu, line 73: similar comment to above, explicitly do `elif self.model_type == "gladia"` and raise ValueError in `else`

jerryjliu (commented):
looks like there's some import errors, i'll fix soon

loader_hub/file/audio/base.py
ravi03071991: sure. will update.

loader_hub/file/audio/base.py
```diff
@@ -17,23 +17,40 @@ class AudioTranscriber(BaseReader):
 
     """
 
-    def __init__(self, *args: Any, model_version: str = "base", **kwargs: Any) -> None:
+    def __init__(self, *args: Any, model_version: str = "base", 
+                model_type: str = "openai_whisper", diarization_max_speakers: str = None,
+                language: str = None, language_behaviour: str = "automatic multiple languages",
+                target_translation_language: str = 'english', gladia_api_key: str = None,
```
jerryjliu, line 23: To make this interface a bit cleaner, could we create a separate dataclass as an AudioConfig? there could be two subclasses, with openai and gladia as the two:

```python
@dataclass
class WhisperAudioConfig(BaseAudioConfig):
    """Whisper audio config."""
    model_version: str


@dataclass
class GladiaAudioConfig(BaseAudioConfig):
    """Gladia Audio Config.""" 
    gladia_api_key: str
    diarization_max_speakers: str
    language: str
    language_behaviour: str
    ...
```

Then have the `__init__` of the AudioTranscriber take in the model_type and an audio_config arg (we could keep model_version too to keep it backwarsd compatible)

jerryjliu:
heads up @ravi03071991 i ended up moving the gladia logic to a separate file, thought it was cleaner to separate the arguments since they are pretty distinct

jerryjliu (approved):

PR #170, opened 2023-04-07
Json data loader

Josh-XT:
Json data loader to directly pass in data received from any REST API that returns JSON format.


jerryjliu:
will take a look soon!

jerryjliu (approved):
we'd probably want to consolidate with the orig json loader in the future

PR #174, opened 2023-04-08
add support for LangChain VectorstoreIndexCreator

mohit:
- Add a `load(self,..)` method to support direct use with LangChain's `VectorstoreIndexCreator.from_loaders([...])` method
- add LanChain usage example


EmanuelCampos:
There's a reason to have it here? instead of only running it after loading the data?

EmanuelCampos:
Closing it for now, due inactivity, feel free to open again

PR #175, opened 2023-04-08
Extend S3Reader

claudenm:
Other cloud providers, such as Linode & DigitalOcean conform to the S3 Api.
This PR adds the ability to select the url to which requests should be sent instead of just to S3.


jerryjliu (approved):
this is great! thanks

PR #181, opened 2023-04-10
Add Zulip Reader

blue-note:


jerryjliu:
sweet! thanks so much

loader_hub/library.json
```diff
@@ -1,474 +1,478 @@
...
+            "pdf",
+            "flat",
+            "flattened"
+        ]
+    },
+    "MilvusReader": {
+        "id": "milvus",
+        "author": "filip-halt"
+    },
+    "ZuilpReader": {
```
jerryjliu, line 474: i trust that this is just reformatting and there's not major changes here

loader_hub/zulip/requirements.txt
```diff
@@ -0,0 +1,2 @@
+zulip
+llama_index
```
jerryjliu, line 2: don't need llama-index as a requirement (since the client to use llamahub is llama_index)

jerryjliu (approved):
thanks!

loader_hub/zulip/base.py
```diff
@@ -0,0 +1,66 @@
+import logging
+from typing import List, Optional
+from datetime import datetime
+import os
+import zulip
```
jerryjliu, line 5: could this be a lazy import within the `__init__.py` file? see other data loaders as an example

mostly because it's breaking tests atm

jerryjliu: @blue-note bumping this! would be good to get this fixed, then we can land

jerryjliu: thanks!

PR #192, opened 2023-04-13
Adding Hubspot loader

ykhli:
- Loads Deals, Contacts, and Companies objects into documents 
- Supports auth token, obtained by creating a private app on Hubspot


jerryjliu:
amazing! will take a look soon

ykhli:
Great! I was actually trying to find a good example for giving the user more out of box ways to structure Documents differently, since different use cases benefit from different Documents structures. 

For example, instead of returning a list of Documents naively, it may be more handy (at least in my own use case) to return a list of Documents split by time or owner for a potentially faster query later. Is this a pattern you have seen elsewhere? or is this part usually handled on the node level?

loader_hub/hubspot/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Usage
+
+Here's an example usage of the HubspotReader.
+
+```python
+from llama_index import download_loader
+import os
+HubspotReader = download_loader('HubspotReader')
+
+reader = HubspotReader("<HUBSPOT_ACCESS_TOKEN">)
```
jerryjliu, line 16: nit: move right bracket inside quote

jerryjliu (approved):
approving with one nit

PR #199, opened 2023-04-17
Add paging for space key and ability to dfs children pages

nikunjy:
1. `space_key` will now page through the confluence API in a limited way as opposed to requesting a very large limit 
2. Ability to DFS children. I believe this is going to be used more often than not.

If you like the children option I can do it for the `label` option as well.

### DIFF
--- a/loader_hub/confluence/base.py
+++ b/loader_hub/confluence/base.py
@@ -34,7 +34,7 @@ class ConfluenceReader(BaseReader):
             from atlassian import Confluence
         except ImportError:
             raise ImportError("`atlassian` package not found, please run `pip install atlassian-python-api`")
-
+        self.confluence: Confluence = None
         if oauth2:
             self.confluence = Confluence(url=base_url, oauth2=oauth2, cloud=True)
         else:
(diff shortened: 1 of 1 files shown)


nikunjy:
if interested in merging please review @emptycrown  / @jerryjliu  

It was useful to me when I was using it.

loader_hub/confluence/base.py
```diff
@@ -92,13 +106,26 @@ def load_data(self, space_key: Optional[str] = None, page_ids: Optional[List[str
...
+            if include_children:
+                page_ids = self._dfs_page(self.confluence, page_ids[0])
             for page_id in page_ids:
                 page = self.confluence.get_page_by_id(page_id=page_id, expand='body.storage.value')
                 doc = self.process_page(page, include_attachments, text_maker)
                 docs.append(doc)
 
         return docs
 
+    def _dfs_page(self, raw_confluence, page_id):
```
nikunjy, line 120: I can combine this and the getting the page itself as well but keeping it this way for now.

jerryjliu:
thanks! will review tomorrow

loader_hub/confluence/base.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, base_url: str = None, oauth2: Optional[Dict] = None) -> None:
             from atlassian import Confluence
         except ImportError:
             raise ImportError("`atlassian` package not found, please run `pip install atlassian-python-api`")
-
+        self.confluence: Confluence = None
```
jerryjliu, line 37: you don't need this line (it's set in the if/else below)

jerryjliu (approved):
amazing!

nikunjy:
Thanks for merging, was traveling so couldn't do it.

PR #201, opened 2023-04-17
Loader for local version of Outlook Calendar

tevslin:
This loader creates a document list from the local contents of an outlook calendar in an .ost or .pst on the user's machine. It deliberately has a superset of the capabilities of the Google Calendar Reader and the same defaults for ease of use by developers who want to support both types of calendar from the same code base. Since only the local calendar is accessed and the loader knows how to find that, no UI or access key is needed.

This functions on Windows only.


tevslin:
I think the problem is that the test was run on Linux and this is windows only. Win32.client gets installed as part the install of pywin32 which is in requirements.text but it is windows specific. There isn't use for this on linux because there is now way to build a local outlook file on linux. I am thinking of a windows-independent version which queries the exchange-hosted calendar file but I think this is a different user case and not usable for those who don't have exchange-hosting of the mail and calendar data. can the tets be run in a Windows environment?

loader_hub/outlook_localcalendar/base.py
```diff
@@ -0,0 +1,108 @@
...
+"""
+Created on Sun Apr 16 12:03:19 2023
+
+@author: tevslin
+"""
+
+
+
+import datetime
+import win32com.client
```
jerryjliu, line 12: you should import this lazily (that's why it is failing tests)

jerryjliu (commented):
one comment about lazy imports! afterwards i'll approve

tevslin:
added lazy load and also graceful fail if not windows by returning empty list. Thanks.

jerryjliu (approved):

PR #211, opened 2023-04-23
Update slack reader

ravi03071991:
Hi,

I have updated the Slack reader due to issues encountered while retrieving messages using the earliest and latest dates. Specifically, I found that errors were occurring because the `conversation_history` did not include these dates. Therefore, we have made changes to address this issue.

Thank you.

### DIFF
--- a/loader_hub/slack/base.py
+++ b/loader_hub/slack/base.py
@@ -116,10 +116,18 @@ class SlackReader(BaseReader):
                 # conversations.history returns the first 100 messages by default
                 # These results are paginated,
                 # see: https://api.slack.com/methods/conversations.history$pagination
-                result = self.client.conversations_history(
-                    channel=channel_id,
-                    cursor=next_cursor,
-                )
+                if self.earliest_date_timestamp is None:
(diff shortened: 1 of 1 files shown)


loader_hub/slack/base.py
```diff
@@ -116,10 +116,18 @@ def _read_channel(self, channel_id: str, reverse_chronological: bool) -> str:
...
                 # These results are paginated,
                 # see: https://api.slack.com/methods/conversations.history$pagination
-                result = self.client.conversations_history(
-                    channel=channel_id,
-                    cursor=next_cursor,
-                )
+                if self.earliest_date_timestamp is None:
+                    result = self.client.conversations_history(
+                        channel=channel_id,
+                        cursor=next_cursor,
```
jerryjliu, line 122: shouldn't this also include latest?

ravi03071991: we are filtering it based on earliest date and so the code is modified accordingly.

jerryjliu (approved):

PR #216, opened 2023-04-27
Update pandas_excel loader

yudataguy:
Based on my own use case, I've updated `pandas_excel` loader to maximize batch loading capability. 
* Load entire excel file with multiple sheets
* Option to load individual sheet within excel file
* Load entire sheet without specify `column_name` 

This will greatly improve productivity when loading large quantity of files, and excel with multiple sheets.

### DIFF
--- a/loader_hub/file/pandas_excel/base.py
+++ b/loader_hub/file/pandas_excel/base.py
@@ -35,7 +35,7 @@ class PandasExcelReader(BaseReader):
         self._pandas_config = pandas_config
 
     def load_data(
-        self, file: Path, extra_info: Optional[Dict] = None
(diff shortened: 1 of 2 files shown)


jerryjliu (approved):
thanks! i think the improvements are worth the breakages in BW compatibility

loader_hub/file/pandas_excel/base.py
```diff
@@ -35,7 +35,7 @@ def __init__(
         self._pandas_config = pandas_config
 
     def load_data(
-        self, file: Path, column_name: str, extra_info: Optional[Dict] = None
+        self, file: Path, sheet_name: str | int | None = None, extra_info: Optional[Dict] = None
```
jerryjliu, line 38: Instead of using | you should do Optional[Union[int, str]]

PR #221, opened 2023-04-29
Add asynchronous web page reader

Hironsan:
- Loads web pages into documents asynchronously
- Supports a maximum number of concurrent requests


jerryjliu (approved):
Thanks, thought about it and i think it makes sense as a separate reader for now

PR #228, opened 2023-05-02
add MetalReader

softboyjimbo:
## MetalReader
https://getmetal.io

ref https://github.com/jerryjliu/llama_index/pull/1784


jerryjliu (commented):
you'd need to add this to the outer library.json file in order for the loader to be registered.

also add metal_sdk to requirements.txt in this folder

jerryjliu (approved):

PR #255, opened 2023-05-12
Expand `body.storage.value` when getting pages from space

aerickson-clt:
With reference to https://github.com/emptycrown/llama-hub/issues/254

 `body.storage.value` needs to be expanded so that `process_page()` does not encounter `KeyError: 'body'`


bdjohnson529:
+1, I also encountered this error, and this fixed it

jerryjliu (approved):
thanks!

PR #261, opened 2023-05-15
Update the google_drive loader to be able to query files by mimeType

badcom:
Added the ability to filter files by the mimeType.

Usage example:
loader.load_data(folder_id=folder_id, mime_types=["application/vnd.google-apps.document"])

The script will still search for the files recursively.

LMK if you need me to make any changes.

### DIFF
--- a/loader_hub/google_drive/base.py
+++ b/loader_hub/google_drive/base.py
@@ -106,7 +106,7 @@ class GoogleDriveReader(BaseReader):
         return creds, drive
 
     def _get_fileids_meta(
-        self, folder_id: Optional(str) = None, file_id: Optional(str) = None, mime_types: Optional(list) = None
+        self, folder_id: Optional[str] = None, file_id: Optional[str] = None, mime_types: Optional[list] = None
     ) -> List[str]:
         """Get file ids present in folder/ file id
         Args:
--- a/loader_hub/google_drive/base.py
+++ b/loader_hub/google_drive/base.py
@@ -106,12 +106,13 @@ class GoogleDriveReader(BaseReader):
         return creds, drive
 
     def _get_fileids_meta(
-        self, folder_id: str = None, file_id: str = None
+        self, folder_id: str = None, file_id: str = None, mime_types: list = None
     ) -> List[str]:
         """Get file ids present in folder/ file id
         Args:
             folder_id: folder id of the folder in google drive.
             file_id: file id of the file in google drive
+            mime_types: the mimeTypes you want to allow e.g.: "application/vnd.google-apps.document"
         Returns:
             metadata: List of metadata of filde ids.
         """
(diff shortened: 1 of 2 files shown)


loader_hub/google_drive/base.py
```diff
@@ -106,12 +106,13 @@ def _get_credentials(self) -> Any:
         return creds, drive
 
     def _get_fileids_meta(
-        self, folder_id: str = None, file_id: str = None
+        self, folder_id: str = None, file_id: str = None, mime_types: list = None
```
jerryjliu, line 109: nit: type these argument names as `Optional[str]`, `Optional[str]`, `Optional[list]` respectively (i know they were incorrect before)

loader_hub/google_drive/base.py
```diff
@@ -122,6 +123,13 @@ def _get_fileids_meta(
             if folder_id:
                 fileids_meta = []
                 query = "'" + folder_id + "' in parents"
+
+                #Add mimeType filter to query
+                if mime_types:
+                    mime_types.append("application/vnd.google-apps.folder") #keep the recursiveness
```
jerryjliu, line 129: it's not obvious to me why you always append this - wouldn't you have duplicate values of `"application/vnd.google-apps.folder"` in mime_types?

loader_hub/google_drive/base.py
```diff
@@ -122,6 +123,13 @@ def _get_fileids_meta(
...
+                    mime_query = " or ".join([f"mimeType='{mime_type}'" for mime_type in mime_types])
+                    query += f" and ({mime_query})"
+
```
jerryjliu, line 132: should mime_types also be passed to _get_fileids_meta below?

jerryjliu (commented):
thanks - some comments

jerryjliu (commented):
thanks - some comments

loader_hub/google_drive/base.py
badcom: Yes, you're right. How about checking if the folder mime type is in the list before adding it?
It needs to be in the list, otherwise the query will skip the subfolders making it non-recursive.

loader_hub/google_drive/base.py
badcom: You're right. Fixed.

loader_hub/google_drive/base.py
badcom: See the change I made to it

loader_hub/google_drive/base.py
badcom: Done!

loader_hub/google_drive/base.py
jerryjliu: yeah that works

jerryjliu (approved):
lgtm, approving 👍 

to fix failing tests you'll want to merge the `main` branch (i can also do in a few hours)

badcom:
> lgtm, approving 👍
> 
> to fix failing tests you'll want to merge the `main` branch (i can also do in a few hours)

Hmmm... I'm not seeing an option to do this =X

PR #263, opened 2023-05-16
Added GraphQL Loader

jexp:
Add a loader for graphql APIs

- uses `gql` graphql library
- configure with uri + headers
- query with graphql query + variables (parameters)
- turn result documents into YAML which is closest to text representation for JSON (open to other suggestions)


loader_hub/graphql/base.py
```diff
@@ -0,0 +1,73 @@
...
+                "`gql` package not found, please run `pip install gql`"
+            )
+        if uri:
+            if uri is None:
+                raise ValueError("`uri` must be provided.")
+            transport = RequestsHTTPTransport(url=uri, headers=headers)
+            self.client = Client(transport=transport, fetch_schema_from_transport=True)
+            
+    def load_data(
+        self, query: str, variables: Dict = {}
```
jerryjliu, line 42: nit: I would type as `Optional[Dict]` and set default value to None here, and then in the function call do `variables = variables or {}`. Setting `{}` in the default params can have unintended side effects

jerryjliu (commented):
looks good to me! one nit

also do you have a demo screenshot or video of using this loader? Would love to help publicize on twitter

jerryjliu:
@jexp quick bump on this - I can also get to it

jexp:
Hi sorry, I missed your previous message. I can change it. No problem.

jerryjliu (approved):

PR #270, opened 2023-05-17
Support to shared drives

badcom:
Add support to Shared Drives


loader_hub/google_drive/base.py
```diff
@@ -132,35 +132,61 @@ def _get_fileids_meta(
...
-                                item["owners"][0]["displayName"],
-                                item["name"],
-                                item["createdTime"],
-                                item["modifiedTime"],
+                        # Check if file doesn't belong to a Shared Drive. "owners" doesn't exist in a Shared Drive
+                        if 'driveId' not in item:
+                            fileids_meta.append(
+                                (
+                                    item["id"],
+                                    item["owners"][0]["displayName"],
```
jerryjliu, line 153: instead of having a different length tuple dependent on shared drive or not, can we just have a boolean variable in `fileids_meta` representing `isShared`? 

having a list of tuples that are of varying lengths is not clean, both for typing purposes as well as code processing purposes

loader_hub/google_drive/base.py
```diff
@@ -214,13 +263,23 @@ def get_metadata(filename):
...
                     final_filepath = self._download_file(fileid, filepath)
-                    metadata[final_filepath] = {
-                        "file id": fileid_meta[0],
-                        "author": fileid_meta[1],
-                        "file name": fileid_meta[2],
-                        "created at": fileid_meta[3],
-                        "modified at": fileid_meta[4],
-                    }
+                    # File is not in a Shared Drive
+                    if len(fileid_meta) == 5:
```
jerryjliu, line 267: if you just have a boolean variable in fileid_meta you don't need this check

loader_hub/google_drive/base.py
```diff
@@ -176,19 +202,42 @@ def _download_file(self, fileid: str, filename: str) -> str:
...
-            if file["mimeType"] in self._mimetypes:
-                download_mimetype = self._mimetypes[file["mimeType"]]["mimetype"]
-                download_extension = self._mimetypes[file["mimeType"]]["extension"]
-                new_filename = filename + download_extension
-                # download file with filename and mimetype
-                file.GetContentFile(new_filename, mimetype=download_mimetype)
-                return new_filename
+            # Get file details
+            service = build("drive", "v3", credentials=self._creds)
+            file = service.files().get(fileId=fileid, supportsAllDrives=True).execute()
```
jerryjliu, line 213: out of curiosity, why do this instead of self._drive.CreateFile?

loader_hub/google_drive/base.py
```diff
@@ -176,19 +202,42 @@ def _download_file(self, fileid: str, filename: str) -> str:
...
-                # download file with filename
-                file.GetContentFile(filename)
-                return filename
+                new_file_name = filename
+
+                # Download file without conversion
+                request = service.files().get_media(fileId=fileid)
+
+            # Download file data
+            file_data = BytesIO()
```
jerryjliu, line 229: why do this instead of file.GetContentFile?

loader_hub/google_drive/base.py
badcom: Because [pydrive](https://github.com/googlearchive/PyDrive) has been deprecated and doesn't support Shared Drives. We might wanna replace pydrive entirely

loader_hub/google_drive/base.py
badcom: Because [pydrive](https://github.com/googlearchive/PyDrive) has been deprecated and doesn't support Shared Drives. We might wanna replace pydrive entirely

loader_hub/google_drive/base.py
badcom: Good idea. See the change I made

loader_hub/google_drive/base.py
jerryjliu: makes sense

loader_hub/google_drive/base.py
```diff
@@ -132,35 +132,67 @@ def _get_fileids_meta(
...
-                            (
-                                item["id"],
-                                item["owners"][0]["displayName"],
-                                item["name"],
-                                item["createdTime"],
-                                item["modifiedTime"],
+                        # Check if file doesn't belong to a Shared Drive. "owners" doesn't exist in a Shared Drive
+                        is_shared_drive = 'driveId' in item
+                        if is_shared_drive:
+                            fileids_meta.append(
```
jerryjliu, line 151: to clarify, my overall point here is that tuples to fileids_meta should be the same length - it's just cleaner when reasoning about typing. if this is a matter of reasoning whether `item["owners"][0]["displayName"]` exists, then you can do something like `display_name = item["owners"][0]["displayName"] if is_shared_drive or None`, and pass that in to the tuple.

badcom: Yes, that makes sense. I've pushed a new commit

jerryjliu (approved):

PR #271, opened 2023-05-18
Adding Docugami loader to llama-hub

tjaffri:
## Adds a data loader for Docugami

Specifically:

1. Adds a data loader that talks to the [Docugami](http://docugami.com) API to download processed documents (scanned pdf, digital native pdf, docx, doc) as semantic XML
2. Parses the semantic XML into chunks, with additional metadata capturing chunk semantics
3. Adds a detailed notebook showing how you can use additional metadata returned by Docugami for more precise Q&A.

Docugami is natively supported in LangChain: https://python.langchain.com/en/latest/integrations/docugami.html

Here is an example of a result that is not possible without the capabilities added by Docugami (from the notebook):

image


loader_hub/docugami/base.py
```diff
@@ -0,0 +1,321 @@
+"""Docugami reader."""
+
+import io
+import os
+import re
+
+from lxml import etree
```
jerryjliu, line 7: would you be able to put this as a lazy import within the function ? otherwise tests will break

jerryjliu (commented):
looks good to me at a high-level, one comment here

loader_hub/docugami/base.py
tjaffri: sure!

PR #272, opened 2023-05-18
Added oAuth to Jira loader

subhrajit2564:
Work done-
1- Added Oauth process to Jira loader and updated the readme accordingly 
2- Fixed a bug where I moved the issues.append outside of the if condition of  "if "parent" in (issue.raw["fields"])". This was causing empty documents to be returned if there was no .raw was found. 

All the changes are tested in my local.


loader_hub/jira/base.py
```diff
@@ -1,75 +1,104 @@
-from typing import List
+from typing import List, Optional, Dict, TypedDict
 
 from llama_index.readers.base import BaseReader
 from llama_index.readers.schema.base import Document
 
+
+class basic_auth(TypedDict):
```
jerryjliu, line 7: can we type name with camel case and not snake case? 

class BasicAuth(TypedDict)

class OAuth(TypedDict):

loader_hub/jira/base.py
```diff
@@ -1,75 +1,104 @@
...
+        Optional oauth:{
+            "cloud_id": "cloud_id",
+            "api_token": "token"
+        }
     """
 
-    def __init__(self, email: str, api_token: str, server_url: str) -> None:
-
+    def __init__(
+        self, basic_auth: Optional[basic_auth] = None, oauth2: Optional[oauth] = None
```
jerryjliu, line 34: this will break backwards compatibility. Can we keep existing variable names (maybe replace type of str with Optional[str] with default kwarg = None), and then add a `basic_auth` and `oauth2` variable? if `basic_auth` or `oauth2` are specified we can do checks to make sure other variable names aren't specified

jerryjliu (commented):
> This was causing empty documents to be returned if there was no .raw was found.

So the assumption here is that if no .raw was found, we should still return a Document right?

loader_hub/jira/base.py
subhrajit2564: Sure, let me add that.

loader_hub/jira/base.py
subhrajit2564: I will do the changes.

jerryjliu (approved):

jerryjliu:
@subhrajit2564 seems like there's a merge conflict, can land after this is resolved

jerryjliu:
@subhrajit2564 quick bump on this

EmanuelCampos:
bump @subhrajit2564

PR #279, opened 2023-05-20
monday.com data reader

monadav:
Adding monday.com data reader based on monday board_id.
The reader requires an api_key as an argument in its constructor and fetches data via monday.com public graphQL API


jerryjliu:
yesss this is going to be super useful! reviewing now

loader_hub/mondaydotcom/README.md
```diff
@@ -0,0 +1,21 @@
...
+MondayReader = download_loader('MondayReader')
+
+reader = MondayReader("<monday_api_token>")
+documents = reader.load_data("<board_id: int>")
+
+```
+
+Check out monday.com API docs - [here](https://developer.monday.com/apps/docs/mondayapi)
+
+This loader is designed to be used as a way to load data into [LlamaIndex](https://github.com/jerryjliu/gpt_index/tree/main/gpt_index) and/or subsequently used as a Tool in a [LangChain](https://github.com/hwchase17/langchain) Agent. See [here](https://github.com/emptycrown/llama-hub/tree/main) for examples.
```
jerryjliu, line 21: can you fix llamaindex link to https://github.com/jerryjliu/llama_index?

loader_hub/mondaydotcom/base.py
```diff
@@ -0,0 +1,89 @@
...
+        api_key (str): monday.com API key.
+    """
+    
+    def __init__(self, api_key: str) -> None:
+        """Initialize monday.com reader."""
+        
+        self.api_key = api_key
+        self.api_url = "https://api.monday.com/v2"
+
+    def parse_item_values(self, cv):
```
jerryjliu, line 22: more of a nit since we don't epxlicitly enforce this yet, but would be nice to explicitly type return values of these functions

jerryjliu (commented):
generally looks good! some nits

jerryjliu (commented):
generally looks good! some nits

jerryjliu (approved):

PR #280, opened 2023-05-20
File reference metadata

anoopshrma:
### Summary
- Replaced `PyPDF2` with `pypdf` library to get page reference.
- Added support for including page reference for pdf file. Added by [Simon](https://github.com/Disiok) on llama-Index.
- Added support for adding file name in metadata for PDF and Docx type document

### Reason to make these changes
This will help in case when the person is handling multiple documents at once. Not only we can provide the response but also the page reference in case of pdf file and also the file name in case of pdf and docx document. This makes it easier to find the exact document.

### DIFF
--- a/loader_hub/file/docx/base.py
+++ b/loader_hub/file/docx/base.py
@@ -17,5 +17,9 @@ class DocxReader(BaseReader):
         import docx2txt
 
         text = docx2txt.process(file)
+        metadata = {"file_name": file.name}
+
+        if extra_info is not None:
+            metadata.update(extra_info)
 
         return [Document(text, extra_info=extra_info)]
--- a/loader_hub/file/docx/base.py
+++ b/loader_hub/file/docx/base.py
@@ -22,4 +22,4 @@ class DocxReader(BaseReader):
         if extra_info is not None:
             metadata.update(extra_info)
 
-        return [Document(text, extra_info=extra_info)]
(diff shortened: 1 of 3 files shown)


jerryjliu (commented):
@anoopshrma i think this is really close. if you're looking at fixing discrepancies between llama-hub and the docsreader in llama-index though you may want to look at making each page a separate Document (see PDFReader in llama-index) 

also happy to land this PR if you'd want to leave that for later, lmk!

anoopshrma:
> @anoopshrma i think this is really close. if you're looking at fixing discrepancies between llama-hub and the docsreader in llama-index though you may want to look at making each page a separate Document (see PDFReader in llama-index)
> 
> also happy to land this PR if you'd want to leave that for later, lmk!

Done! Now each page is added as a separate document @jerryjliu

jerryjliu:
> > @anoopshrma i think this is really close. if you're looking at fixing discrepancies between llama-hub and the docsreader in llama-index though you may want to look at making each page a separate Document (see PDFReader in llama-index)
> > also happy to land this PR if you'd want to leave that for later, lmk!
> 
> Done! Now each page is added as a separate document @jerryjliu

thanks!

jerryjliu (approved):

jerryjliu (approved):

PR #281, opened 2023-05-20
Add Azure Storage blob loader

rivms:
Add a loader for Azure Storage.

- Passes through to Azure Storage blob sdk 
- Supports SAS and Azure AD authentication


loader_hub/azstorage_blob/base.py
```diff
@@ -0,0 +1,100 @@
...
+
+"""
+import logging
+import tempfile
+import time
+import math
+from pathlib import Path
+from typing import Any, List, Optional, Union, Dict
+
+from azure.core.credentials import AzureNamedKeyCredential, AzureSasCredential, TokenCredential
```
jerryjliu, line 13: you'll need to do lazy imports for tests to pass

loader_hub/azstorage_blob/base.py
```diff
@@ -0,0 +1,100 @@
...
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+
+logger = logging.getLogger(__name__)
+
+class AzStorageBlobReader(BaseReader):
+    """General reader for any Azure Storage Blob file or directory.
+
+    Args:
+        service_name
```
jerryjliu, line 26: i would fill out these arguments more comprehensively in docstring

PR #284, opened 2023-05-21
snscrape loader for twitter

Smyja:
This loader allows for fetching tweets from a user's profile.


jerryjliu (commented):
out of curiosity is there a way to design a general SnscrapeReader that can read in any social media source or is that hard

Smyja:
> out of curiosity is there a way to design a general SnscrapeReader that can read in any social media source or is that hard

Yes, it would be possible. But twitter is the only one that seems to be working for snscrape.

jerryjliu (approved):

PR #285, opened 2023-05-22
Fix base.py for flat_pdf

bhavishpahwa:
Fixed the bug that was coming up due to this commit and ImageReader output being changed due to this commit:- https://github.com/emptycrown/llama-hub/commit/a37df8c22194b2e15ac509aeeb86a2680c9556bf#diff-d4cdba126dd16c86090107c1c66621a520d20d8962f7f4dd461146cedbb1135d.

Also for future proofing can llama-hub also be versioned so that it can be more robust for making end to end prod systems. Changes in llama-hub can affect document/pdf/input digestion process while using it with llamaindex and break pipelines. Any thoughts on this @jerryjliu


jerryjliu (approved):

jerryjliu:
@bhavishpahwa yeah it's a good point...we're working on making llamahub a full pypi package (with versioning). hope to have this soon!

PR #292, opened 2023-05-25
Added JoplinReader

alondmnt:
[Joplin](https://joplinapp.org/) is an open source note-taking app.

Joplin has a [REST API](https://joplinapp.org/api/references/rest_api/) for accessing its local database. The proposed `JoplinReader` uses the API to retrieve all notes in the database and their metadata. Joplin needs to be installed and running locally, and an access token is required.

### DIFF
--- a/loader_hub/joplin/base.py
+++ b/loader_hub/joplin/base.py
@@ -10,7 +10,6 @@ import os
 from typing import Any, Iterator, List, Optional
 import urllib
 
-from langchain.docstore.document import Document as LCDocument
 from llama_index.readers.base import BaseReader
 from llama_index.readers.schema.base import Document
 
--- a/loader_hub/joplin/base.py
+++ b/loader_hub/joplin/base.py
@@ -25,7 +25,8 @@ class JoplinReader(BaseReader):
     Web Clipper enabled (look for "Web Clipper" in the app settings).
 
     To get the access token, you need to go to the Web Clipper options and
-    under "Advanced Options" you will find the access token.
+    under "Advanced Options" you will find the access token. You may provide
+    it as an argument or set the JOPLIN_ACCESS_TOKEN environment variable.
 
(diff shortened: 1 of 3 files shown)


loader_hub/joplin/base.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def lazy_load(self) -> Iterator[Document]:
+        yield from self._get_notes()
+
+    def load_data(self) -> List[Document]:
+        return list(self.lazy_load())
+
+    def load_langchain_documents(self, **load_kwargs: Any) -> List[LCDocument]:
+        """Load data in LangChain document format."""
+        docs = self.load_data(**load_kwargs)
```
jerryjliu, line 104: don't think you need to override this method

loader_hub/joplin/base.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class JoplinReader(BaseReader):
+    """
+    Reader that fetches notes from Joplin.
+
+    In order to use this reader, you need to have Joplin running with the
+    Web Clipper enabled (look for "Web Clipper" in the app settings).
+
+    To get the access token, you need to go to the Web Clipper options and
```
jerryjliu, line 27: would be nice to put info about JOPLIN_ACCESS_TOKEN in the docstring

also more generally document the args

jerryjliu (commented):
looks pretty good! some minor comments

loader_hub/joplin/base.py
alondmnt: right, thanks! saw it in one of the readers I used as reference (ObsidianReader), but now I see that BaseReader already implements it. removing.

loader_hub/joplin/base.py
alondmnt: good idea, adding.

jerryjliu (approved):

PR #293, opened 2023-05-25
Here i am adding the OSM(open street map) loader

carrotpy:
https://github.com/emptycrown/llama-hub/assets/48360598/8dda616c-8281-4ce7-af53-35dcd2984eb6


loader_hub/library.json
```diff
@@ -489,6 +489,11 @@
     "author": "iamadhee",
     "keywords": ["weather","openweather"]
   },
+    "OpenMap": {
+    "id": "osmmap",
+    "author": "carrotpy",
+    "keywords": ["open maps","maps","open street maps","overpass api"]
```
iamadhee, line 495: Add keywords `maps, geo` as well

carrotpy: Thanks @iamadhee  added the same

PR #318, opened 2023-06-13
refactor: add custom path for download_loader to enable functionality…

jack-hughes:
… on AWS Lambda

AWS Lambda only has ephemeral storage in the /tmp directory. This means whenever attempting to use the S3Reader in conjunction with Lambda, it fails on this line.


jerryjliu (commented):
nit: if we're going to change this, i'd prefer to have an argument in the `__init__.py` called `custom_reader_path` - will only be used if `ImportError` is triggered

speaking of which, any reason why you can't do `from llama_hub.s3.base import S3Reader` by `pip install llama-hub` as a package?

logan-markewich:
@jerryjliu one nice thing about download loader is it install dependencies for you. Tbh I wish importing from the package did that too

EmanuelCampos:
I'd prefer to have a custom_reader_path in `__init__py` as well, but instead of waiting for an ImportError we would use it, there's other reasons to use a custom path

EmanuelCampos:
@logan-markewich @jerryjliu any thoughts on it? added a custom_reader_path

logan-markewich (approved):

PR #323, opened 2023-06-21
support Confluence personal access token

alexlopashev:
Existing support for Confluence basic auth lacks popular option of providing personal access token.

This PR changes behavior of CONFLUENCE_API_TOKEN environment variable, which is used to provide user's password, for this purpose new environment variable CONFLUENCE_PASSWORD is now supported.

Additionally, inner Confluence API is able to support self-hosted (not Atlassian Cloud) instances just fine, for that I've added `cloud` bool argument, defaults to True to preserve current behavior.


jerryjliu (approved):
this is a breaking change for existing users but i'm ok with landing this for now

jerryjliu:
@alexlopashev looks like this is failing tests, would you be able to help look into fixing?

alexlopashev:
@jerryjliu apologies for the delay, tests were adjusted and passing locally.

> this is a breaking change for existing users but i'm ok with landing this for now
Yes, it is a breaking change, but imho it's worth to update environment variables names to prevent confusion I had when tried to use access tokens

PR #324, opened 2023-06-21
Change Docugami Data Loader Namespace

kenzie28:
Update the artifact name of the xml file and the namespaces. Also fixed links in the readme file.

Co-authored with @tjaffri


kenzie28:
Also would like to ask how can we test our notebook using local changes since importing llama gets the remote version?

jerryjliu:
@kenzie28 now that llama_hub is a proper package, it should be easier? 

just clone repo, do `poetry install`, and run your notebook with `from llama_hub .. import ...`

jerryjliu (approved):

PR #330, opened 2023-06-23
confluence.load_data new features, bug fixes, tests

aerickson-clt:
A number of enhancements to the confluence `load_data` function, as follows:

- Add `retrying` as dependency.
- GET requests to Confluence API are done with `retrying` to reduce chances of one of them killing batches.
- Fix bug(s)s in paged GET requests. (see Additional Information).
- Implement paged GET requests for cql and labels.
- Add more argument validation.
- Deprecate `limit` parameter in favour of unambiguous `max_num_results` parameter.
- Implement `max_num_results` for every retrieval method (cql, label, page_ids, include_child_pages, `space_key`)
- Make the different retrieval methods mutually exclusive (I can't think of when you would want the union of different retrieval methods, and that complicates the use of `max_num_results`, plus it's easy to call it once for each retrieval method you are interested in).
- Refactored to remove repeated code.
- Added many tests, including for DFS, and both paging methods, plus the additional arg validation.
- Added documentation for the changed behaviour.

There are some breaking changes here, but I think it's unlikely that anyone is relying on the confluence hub in production, and it should continue to work fine in demos for every use case except multiple retrieval methods.

## Additional Information

Confluence API uses start and limit parameters for getting batches of content, but usually there is also an (unknown) server limit (50,200,500?).  We also may want to bound the number of results if we are trying to test a request that can get many results.  To avoid confusion between these different types of "limits", `load_data` offers `max_num_results`, and always tries to get as many results in one request as possible, but then inspects the number of results returned before deciding whether to request more with a new `start` and a new `limit=max_num_remaining`.

Previously, paged requests were not implemented for label and cql queries, and now they are.  Furthermore, previously it was possible to get fewer results than those requested via the now deprecated `limit` parameter if the `limit` parameter was greater than the (usually unknown) server limit.

### DIFF
--- a/llama_hub/confluence/README.md
+++ b/llama_hub/confluence/README.md
@@ -9,10 +9,20 @@ For more on authenticating using OAuth 2.0, checkout:
 - https://atlassian-python-api.readthedocs.io/index.html
 - https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/
 
-User then specify a list of `page_ids` and/or `space_key` to load in the corresponding pages into Document objects, if 
-both are specified the union of both sets will be returned. User can also specify a boolean `include_attachments` to 
+Confluence pages are obtained through one of 4 four mutually exclusive ways:
+
+1. `page_ids`: Load all pages from a list of page ids
+2. `space_key`: Load all pages from a space
+3. `label`: Load all pages with a given label
+4. `cql`: Load all pages that match a given CQL query (Confluence Query Language https://developer.atlassian.com/cloud/confluence/advanced-searching-using-cql/ ).
+
+When `page_ids` is specified, `include_children` will cause the loader to also load all descendent pages.
+When `space_key` is specified, `page_status` further specifies the status of pages to load.
+
+User can also specify a boolean `include_attachments` to 
 include attachments, this is set to `False` by default, if set to `True` all attachments will be downloaded and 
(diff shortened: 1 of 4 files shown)


llama_hub/confluence/README.md
```diff
@@ -23,7 +37,7 @@ Here's an example usage of the ConfluenceReader.
 ```python
 from llama_index import download_loader
 
-ConfluenceReader = download_loader('ConfluenceReader')
+ConfluenceReader = download_loader('ConfluenceReader', refresh_cache=True)
```
jerryjliu, line 40: since you're in the process of landing this change, you could also try updating our default usage to 

`from llama_hub.confluence.base import ConfluenceReader` :)

jerryjliu (approved):
high-level looks good to me 

i didn't rigorously verify correctness, main things i looked for:
- BW compat with existing function signature
- proper documentation of changes in readme/function docstring
- tests

jerryjliu:
may have to remerge upstream main + fix tests though

aerickson-clt:
> may have to remerge upstream main + fix tests though

@jerryjliu should I do that or were you making a note for someone here?

jerryjliu:
@aerickson-clt yeah if you're able to - let me know if you run into issues though

aerickson-clt:
@jerryjliu I have have updated the example usage, merged upstream, fixed tests by adding poetry dependency.

jerryjliu:
awesome work @aerickson-clt !

jerryjliu (approved):

PR #332, opened 2023-06-24
feat: add document reader for GH issues

moncho:
The reader retrieves issues from a repo a converts them into documents. Documents text is the concatenation of the issue title and body, document Id is the issue number. Further information about the issue is added to the document extra_info field.

Issues can be filtered by their state (the default being open) and by label.

This is not part of the existing  GithubRepositoryReader because that loader works with repositories content, while this one works with repositories issues. As an alternative, the existing loader could be moved to a subdirectory under github_repo and this new loader could also be moved to under there, so there are separated but under the same root (github_repo), similar to `file` loaders. I guess this approach would break existing code, which is usually not good :)


llama_hub/github_repo_issues/README.md
```diff
@@ -0,0 +1,77 @@
...
+https://github.com/jerryjliu/llama_index, use `owner = jerryjliu` and `repo = llama_index`.
+
+```shell
+export GITHUB_TOKEN='...'
+```
+
+```python
+import os
+
+from llama_hub.github_repo import GitHubRepositoryIssuesReader, GitHubIssuesClient
```
jerryjliu, line 20: github_repo_issues you mean?

llama_hub/github_repo_issues/README.md
```diff
@@ -0,0 +1,77 @@
...
+```shell
+export OPENAI_API_KEY='...'
+export GITHUB_TOKEN='...'
+```
+
+```python
+import pickle
+import os
+
+from llama_index import download_loader, GPTVectorStoreIndex
```
jerryjliu, line 50: nit: it's just called a VectorStoreIndex now (GPT prefix is deprecated)

jerryjliu (changes requested):
this looks great!! 

can you add an entry to libary.json? then it'll show up in the UI

llama_hub/github_repo_issues/README.md
moncho: Yes, that is what I mean :)

moncho: Fixed!

llama_hub/github_repo_issues/README.md
moncho: Changed

jerryjliu (approved):

llama_hub/library.json
```diff
@@ -540,5 +540,15 @@
   "KibelaReader": {
     "id": "kibela",
     "author": "higebu"
+  },
+  "GitHubIssuesReader": {
```
jerryjliu, line 544: Sorry should've clarified, this needs to match the class name: GitHubRepositoryIssuesReader

moncho: Thanks for clarification, fixed!

PR #336, opened 2023-06-26
Update readme files for reference

selamanse:
This change affects the readme files of asyncwebreader and sitemapreader downloaders

- Updating readme for correct usage with llama_hub as pypi package.
- Adding reference to old usage


jerryjliu (approved):
nice thanks!

PR #337, opened 2023-06-26
Add extra info to web

selamanse:
This change adds extra_info to the generated documents of AsyncWebReader

- Source information (the url called) will be added to extra_info of Document class
- deduplication feature will sort out exact same urls in given urls list
- fail on error feature will handle status codes != 200 either with raising an error or logging a warning
- added tests which introduce pytest_httpserver to simulate different situations for different urls

### DIFF
--- a/llama_hub/web/async_web/base.py
+++ b/llama_hub/web/async_web/base.py
@@ -1,9 +1,12 @@
+import logging
 import asyncio
 from typing import List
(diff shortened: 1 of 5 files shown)


jerryjliu (commented):
thanks @selamanse ! would you be able to rebase on top of the latest changes?

jerryjliu (approved):

PR #338, opened 2023-06-27
update document usage

logan-markewich:
Attempt number 2 of updating document usage -> with the recent patch in llama-index, this change is much smaller

Just need to ensure every document has a kwarg for the text field


jerryjliu (approved):
i'm assuming this works lol

PR #343, opened 2023-06-28
ensure metadata always valid

logan-markewich:
Since the new pydantic document objects require extra_info to not be `None`, ensure our loaders also follow this idiom


logan-markewich:
also I ran black to format some stuff lol

jerryjliu (approved):

PR #368, opened 2023-07-09
fix bug where messages are returned without getting message data

superpan:


superpan:
@jerryjliu , this fixes a bug I introduced with the earlier commit.  It was returning the messages without fetching message data.

jerryjliu (approved):

PR #370, opened 2023-07-10
Add Google Keep Reader

pycui:
Reads notes from Google Keep. 
Note Google Keep's API only works if you have (1) Enterprise account (2) use Service Account to authenticate (3) a Domain-wide Delgation enabled for the service account with the Google Keep scopes - which requires a Google Workspace super admin to set up. Making it less useful for normal users.


jerryjliu:
amazing! will take a look tomorrow @pycui

pycui:
@jerryjliu maybe holding off a bit - after re-thinking about the limitations of current Google Keep API, I think it's not practical for most of users of Keep (the API frankly is not as productionized as Google Docs or Drives) . Would we be open to use "unofficial" API libraries like https://github.com/kiwiz/gkeepapi which actually works better for normal user?

llama_hub/google_keep/README.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+## Examples
+
+This loader is designed to be used as a way to load data into [LlamaIndex](https://github.com/jerryjliu/llama_index/tree/main/llama_index) and/or subsequently used as a Tool in a [LangChain](https://github.com/hwchase17/langchain) Agent.
+
+### LlamaIndex
+
+```python
+from llama_index import GPTVectorStoreIndex, download_loader
```
jerryjliu, line 28: nit: *VectorStoreIndex (gpt prefix is deprecated)

jerryjliu (commented):
> @jerryjliu maybe holding off a bit - after re-thinking about the limitations of current Google Keep API, I think it's not practical for most of users of Keep (the API frankly is not as productionized as Google Docs or Drives) . Would we be open to use "unofficial" API libraries (like https://github.com/kiwiz/gkeepapi which actually works better for normal user) in the hub?

In general yeah we'd be open to unofficial library support. The goal of llamahub is really to give users a diversity of choices in getting started (hence why we have 7+ different pdf loaders), so I'd just pick one to start with. honestly i'm down to review/land this if it looks ok (as long as it works)

pycui:
> In general yeah we'd be open to unofficial library support. The goal of llamahub is really to give users a diversity of choices in getting started (hence why we have 7+ different pdf loaders), so I'd just pick one to start with. honestly i'm down to review/land this if it looks ok (as long as it works)

Sounds good I will switch to [gkeepapi](https://github.com/kiwiz/gkeepapi) API given it works better for most. Will rerequest review once ready.

jerryjliu:
sounds good, thanks!

llama_hub/google_keep/README.md
pycui: Done.

llama_hub/google_keep/README.md
```diff
@@ -0,0 +1,74 @@
...
+This loader uses the (unofficial) gkeepapi library. Google Keep does provide an official API, however in order to use it, (1) your account has to be an Enterprise (Google Workspace) account (2) you will need to generate a service account to autheticate with Google Keep API (3) you will need to enable Domain-wide Delegation to enable the service account with Google Read API scopes. See [here](https://issuetracker.google.com/issues/210500028) for details. Thus I believe gkeepapi is actually more practical and useful for the majority of the users.
+
+To use gkeepapi, you will need to login with username and a password. I highly recommend using a (one-off) App Password over using your own password. You can find how to generate App Password at [here](https://support.google.com/accounts/answer/185833?hl=en). The username and password should be saved at a `keep_credentials.json` file, with `username` and `password` being keys. It's recommended you delete the App Password once you no longer need it.
+
+## Usage
+
+To use this loader, you simply need to pass in an array of Google Keep IDs.
+
+```python
+from llama_index import download_loader
```
jerryjliu, line 14: let's do `from llama_hub.google_keep.base import GoogleKeepReader` (we try to do that by default now)

llama_hub/google_keep/base.py
```diff
@@ -0,0 +1,79 @@
+"""(Unofficial) Google Keep reader using gkeepapi."""
+
+import os
+import gkeepapi
```
jerryjliu, line 4: currently we do lazy imports for tests to pass (not the best eng design decision, it might be worth thinking about a better way to do tests/linting even with optional imports

jerryjliu (commented):
thanks! some comments. also not sure why poetry.lock is modified

jerryjliu (approved):

pycui:
@jerryjliu can you or a maintainer help merge it? Thanks!

jerryjliu (approved):

PR #382, opened 2023-07-13
Code Interpreter tool spec

ajhofmann:
A few things I'm not totally satisfied with:
- I think LlamaIndex probably has an abstraction to do what I'm doing here: https://github.com/emptycrown/llama-hub/pull/382/files#diff-a48282449ce1ed82ccaaea1d9c60d835d2e3230021345e6ee2684c6ee08710a0R42-R53
- Essentially just feeding in the chat history so far to allow the code interpreter tool to iteratively complete tasks
- Unless I prime the agent to use the tool, (use generate_and_execute for the following tasks), it will try to directly invoke python
- By default the tool is initializing a OpenAI LLM to run under the hood, but I think we should be platform agnostic by default, is there a good way to do that?
https://github.com/emptycrown/llama-hub/pull/382/files#diff-a48282449ce1ed82ccaaea1d9c60d835d2e3230021345e6ee2684c6ee08710a0R34


llama_hub/tools/code_interpreter/base.py
```diff
@@ -0,0 +1,81 @@
...
+
+    By default, the only response returned by the code is through stdout. Use print statements heavily.
+
+    Any code that is passed in as a prompt is NOT executed.
+
+    If you are requested to generate a graph, and want to use something like plt.show(), DONT. Instead save the output to a file.
+"""
+
+
+class CodeInterpreterToolSpec(BaseToolSpec):
```
Disiok, line 27: Isn't this basically an agent?

If we add LLM calling, etc, what's the conceptual difference between a tool and an agent?

Disiok: To me. this is a `PythonCodeAgent`, and a `RouterAgent`. 

the `PythonCodeAgent` has specialized prompts to specialize the agent for code generation, and uses a code interpreter as a tool.

llama_hub/tools/code_interpreter/base.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def format_chat(self, prompt: ChatMessage) -> List[ChatMessage]:
+        return [self._initial_prompt, *self._history(), prompt]
+
+    def generate_code(self, prompt: str) -> str:
+        formatted_prompt = ChatMessage(role=MessageRole.USER, content=prompt)
+        response = self._llm.chat(self.format_chat(formatted_prompt))
+        self._prompts.append(formatted_prompt)
+        self._responses.append(ChatMessage(role=MessageRole.ASSISTANT, content=response.message.content))
+        return response.message.content
```
Disiok, line 53: I think this can be represented via the `chat` interface on an agent/chat engine?

llama_hub/tools/code_interpreter/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Usage
+
+Here's an example usage of the CodeInterpreterToolSpec.
+
+```python
+from llama_hub.tools.tool_spec.code_interpreter.base import CodeInterpreterToolSpec
+from llama_index.agent import OpenAIAgent
+
+code_spec = CodeInterpreterToolSpec()
```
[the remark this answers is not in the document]
Disiok: I think might be easier to model this as a _specialized agent_, and then have outer agent calling this agent?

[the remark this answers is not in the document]
Disiok: cc @jerryjliu

ajhofmann:
Thank you for the comments, going to make a bit more of a substantial rework to this PR, closing it in favour of other branches.

PR #385, opened 2023-07-14
Add SingleStoreReader to load documents from a SingleStoreDB

apeng-singlestore:
Loads documents from a singlestore table that has been populated with text and embedding columns.


jerryjliu (commented):
thanks! could we not modify the poetry.lock file, and instead add a requirements.txt file in `llama_hub/singlestore` instead?

PR #386, opened 2023-07-14
Adding functionality for AsanaReader

saranggupta:
Hi! I am an engineer at Asana and I would like to add a couple of functionalities to the AsanaReader. The current implementation only takes the `workspace_id` as input, and returns the list of all tasks. However, oftentimes users just need the ability to extract the task information from one particular project.

As such I have implemented, a functionality that allows user to either enter a `workspace_id` or `project_id` to fetch all relevant tasks. Furthermore, I have also added a few more fields from the task that could be useful as part of the retrieved task data.


llama_hub/asana/base.py
```diff
@@ -19,39 +16,61 @@ def __init__(self, asana_token: str) -> None:
 
         self.client = asana.Client.access_token(asana_token)
 
-    def load_data(self, workspace_id: str) -> List[Document]:
+    def load_data(self, id: str, id_type: str) -> List[Document]:
```
Disiok, line 19: curious: why not just have separate `workspace_id` and `project_id`. It's a lot more explicit, and more backward compatible.

jerryjliu (commented):
@saranggupta super sorry, had a review pending but realized i forgot to submit

EmanuelCampos:
bump @saranggupta

PR #389, opened 2023-07-15
Code interpreter Tool.

ajhofmann:
Reworking the idea of a code-interpreter to a bit more of a simple example to start. This is simply a tool to evaluate python scripts, with access to imports and file system.

The tool requires a bit more prompting on the user end to get working compared to having an Agent that writes the code, however I was able to get it working with an example to read a CSV and create mat plot graphs from it.

By passing stdout and stderr to the Agent it's able to do some level of error recovery as displayed in the notebook.


llama_hub/tools/code_interpreter/base.py
```diff
@@ -0,0 +1,37 @@
...
+class CodeInterpreterToolSpec(BaseToolSpec):
+    """Code Interpreter tool spec."""
+
+    spec_functions = ["code_interpreter"]
+
+    def __init__(
+        self,
+        llm_cls: Optional[LLM] = OpenAI,
+        llm_args: Optional[dict] = {}
+    ) -> None:
```
jerryjliu, line 21: why do we have an `__init__`

jerryjliu (approved):
definitely a lot cleaner than the previous iteration - i'm down to get thi in and get feedback

llama_hub/tools/code_interpreter/base.py
ajhofmann: Oops leftover code, cleaned up!

PR #390, opened 2023-07-15
Updating zapier tool to use parameters

ajhofmann:
Updating the Zapier tool to use the params in addition to the instructions. Also made a bit more of a comprehensive example using the GitHub integrations.

### DIFF
--- a/llama_hub/tools/zapier/base.py
+++ b/llama_hub/tools/zapier/base.py
@@ -63,6 +63,8 @@ class ZapierToolSpec(BaseToolSpec):
         Args:
             id (str): The id of the zapier action to call
             params (Optional[dict]): The instructions and other values instructing the action to be taken
+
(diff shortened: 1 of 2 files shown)


llama_hub/tools/zapier/base.py
```diff
@@ -39,20 +53,21 @@ def list_actions(self):
...
 
+        The action being called must have an id obtained from list_actions. If the action is not exposed it can be here: 
         Args:
             id (str): The id of the zapier action to call
-            instructions (str): The natural language instruction to pass to Zapier
-
+            params (Optional[dict]): The instructions and other values instructing the action to be taken
         """
 
         response = requests.post(
```
jerryjliu, line 68: for future iterations maybe we could think about error handling (if we can find a failure case)

jerryjliu (approved):
this is cool!

Disiok:
I think we can just land this and continue iterating in future PRs.

PR #391, opened 2023-07-16
Update llama_hub/file/pandas_excel/base.py

kevinlu1248:
Copied from https://github.com/kevinlu1248/llama-hub/pull/3 to fix https://github.com/jerryjliu/llama_index/issues/6203.

### Description
This PR fixes two issues with the `PandasExcelReader` class in the `llama_hub/file/pandas_excel/base.py` file.

1. The `load_data` method did not accept a `pandas_config` argument, causing a `TypeError` when trying to pass this argument. This PR modifies the `load_data` method to accept the `pandas_config` argument.

2. The `__init__` method of the `PandasExcelReader` class did not correctly initialize the `_row_joiner` attribute, causing an `AttributeError` when trying to access this attribute. This PR ensures that the `_row_joiner` attribute is correctly initialized in the `__init__` method.

### Changes Made
- Modified the `load_data` method in the `PandasExcelReader` class to accept a `pandas_config` argument.
- Correctly initialized the `_row_joiner` attribute in the `__init__` method of the `PandasExcelReader` class.

### Related Issue
This PR addresses the issue [#1](https://github.com/kevinlu1248/llama-hub/issues/1).

### DIFF
--- a/llama_hub/file/pandas_excel/base.py
+++ b/llama_hub/file/pandas_excel/base.py
@@ -32,7 +32,7 @@ class PandasExcelReader(BaseReader):
         row_joiner: str = "\n",
         **kwargs: Any
     ) -> None:
-        '''Init params.'''
+        """Init params."""
         super().__init__(*args, **kwargs)
         self._pandas_config = pandas_config
         self._concat_rows = concat_rows
--- a/llama_hub/file/pandas_excel/base.py
+++ b/llama_hub/file/pandas_excel/base.py
@@ -27,25 +27,27 @@ class PandasExcelReader(BaseReader):
     def __init__(
         self,
         *args: Any,
-        pandas_config: dict = {},
+        pandas_config: Optional[dict] = None,
         concat_rows: bool = True,
         row_joiner: str = "\n",
         **kwargs: Any
     ) -> None:
         """Init params."""
         super().__init__(*args, **kwargs)
-        self._pandas_config = pandas_config
+        self._pandas_config = pandas_config or {}
         self._concat_rows = concat_rows
         self._row_joiner = row_joiner if row_joiner else "\n"
+        self._pandas_config = pandas_config or {}
 
     def load_data(
         self,
         file: Path,
-        pandas_config: dict = {},
+        pandas_config: Optional[dict] = None,
         include_sheetname: bool = False,
         sheet_name: Optional[Union[str, int]] = None,
         extra_info: Optional[Dict] = None,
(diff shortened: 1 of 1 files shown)


llama_hub/file/pandas_excel/base.py
```diff
@@ -36,11 +36,12 @@ def __init__(
         super().__init__(*args, **kwargs)
         self._pandas_config = pandas_config
         self._concat_rows = concat_rows
-        self._row_joiner = row_joiner
+        self._row_joiner = row_joiner if row_joiner else "\n"
 
     def load_data(
         self,
         file: Path,
+        pandas_config: dict = {},
```
jerryjliu, line 44: this config arg is never used

also it's not safe to initialize to `{}` in fn signature, conventionally what we do is do `pandas_config: Optional[dict] = None`, and then in the function def do something like `pandas_config = pandas_config or {}`

llama_hub/file/pandas_excel/base.py
```diff
@@ -36,11 +36,12 @@ def __init__(
         super().__init__(*args, **kwargs)
         self._pandas_config = pandas_config
         self._concat_rows = concat_rows
-        self._row_joiner = row_joiner
+        self._row_joiner = row_joiner if row_joiner else "\n"
```
jerryjliu, line 39: is this only the case is row_joiner is explicitly specified as None? wondering how this fix helps

kevinlu1248: Don't think I've found anywhere else yet and it seems to get around https://github.com/jerryjliu/llama_index/issues/6203, unless the example's just wrong

llama_hub/file/pandas_excel/base.py
kevinlu1248: Got Sweep to match the suggestions

llama_hub/file/pandas_excel/base.py
jerryjliu: i think the example's just wrong

jerryjliu: since this is a minor non-breaking nit, think i'm going to merge and fix later

jerryjliu (approved):

kevinlu1248:
Hey @jerryjliu, most of this PR was generated by Sweep. Can we try installing Sweep for this repo & llama-index? Installation instructions at https://github.com/sweepai/sweep#-getting-started

PR #393, opened 2023-07-16
mbox: allow custom, stable document id

rc9000:
* via function passed in `id_fn`, eg. `MboxReader(id_fn=lambda msg: md5(msg.encode()).hexdigest())`
 * overrides UUID-based default from Document

### DIFF
--- a/llama_hub/file/mbox/base.py
+++ b/llama_hub/file/mbox/base.py
@@ -113,7 +113,7 @@ class MboxReader(BaseReader):
         for msg in content:
             d = Document(text=msg, extra_info=extra_info or {})
             if self.id_fn:
-                 d.id_ = self.id_fn(msg)
+                 d.doc_id = self.id_fn(msg)
             docs.append(d)
 
         return docs
--- a/llama_hub/file/mbox/base.py
+++ b/llama_hub/file/mbox/base.py
@@ -4,7 +4,7 @@ Contains simple parser for mbox files.
 
 """
 from pathlib import Path
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 from llama_index.readers.base import BaseReader
 from llama_index.readers.schema.base import Document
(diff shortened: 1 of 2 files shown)


llama_hub/file/mbox/base.py
```diff
@@ -109,5 +111,9 @@ def load_data(
         docs: List[Document] = []
         content = self.parse_file(file)
         for msg in content:
-            docs.append(Document(text=msg, extra_info=extra_info or {}))
+            d = Document(text=msg, extra_info=extra_info or {})
+            if self.id_fn:
+                 d.id_ = self.id_fn(msg)
```
jerryjliu, line 116: canonical way to set doc_id is `d.doc_id = ...`.

jerryjliu (commented):
out of curiosity is `msg` a useful argument to pass into `id_fn`? It seems to me that for a lot of use cases you'd need to do a bunch of hacky string parsing to extract out a usable id. 

another option is to modify parse_file, extract a more structured dict (not just the formatted message string, but also a dict containing `_date`, `_from`, `_to_` keys etc.)

llama_hub/file/mbox/base.py
rc9000: Thanks fixed! I think I had this from [here](https://github.com/jerryjliu/llama_index/blob/main/docs/core_modules/data_modules/documents_and_nodes/usage_documents.md?plain=1#L85)

rc9000:
> out of curiosity is `msg` a useful argument to pass into `id_fn`?

I got good and fast results with `id_fn=lambda msg: md5(msg[:200].encode()).hexdigest()`,  since all the identifying information is in the Date:/From:/To: bits which are at the beginning in the default message_format. I put this example in the README now. 

But yes I was also considering passing a dict to id_fn, if you like it a lot more I could also do that.

jerryjliu (approved):

PR #405, opened 2023-07-19
Adding ChatGPT plugin tool

ajhofmann:
A tool that allows you to easily load a ChatGPT Plugin from a manifest file, leveraging the OpenAPI tool spec and Requests tool specs.


llama_hub/tools/chatgpt_plugin/base.py
```diff
@@ -0,0 +1,54 @@
...
+import yaml
+
+class ChatGPTPluginToolSpec(BaseToolSpec):
+    """ChatGPT Plugin Tool
+
+    This tool leverages the OpenAPI tool spec to automatically load ChatGPT 
+    plugins from a manifest file
+    """
+
+    spec_functions = ['describe_plugin']
```
jerryjliu, line 20: hmm i think the tool spec interface is a bit confusing now that we're overriding `to_tool_list`. a bit hard to tell now which tools are actually getting returned to the agent, and the `spec_functions` don't quite match anymore

jerryjliu: do you have thoughts on how to simplify this?

jerryjliu: also, unrelated maybe to this PR but out of curiosity why does the openapi tool take in a pre-existing dict instead of loading the openapi spec as part of the function

ajhofmann: Yeah I see what you mean, I think if we wanted to make the interface more intuitive from the code we could:

- Require the user to import and pass the Requests tool spec, similar to what OpenAPI tool spec requires right now.
- have ChatGPT plugin export its own load_openapi_spec that calls the internally wrapped OpenAPIToolSpec, and add it to spec functions

With that we wouldn't need to override to_tool_list and make it potentially a bit easier for users to customize request headers. I think this seems a bit more straight forward, I'll put together some code.

ajhofmann: For the openapi tool taking a spec, my thinking was it allows users to load a spec from url or local file, but we could also update the interface to accept a file path or url and load the spec from there. I can't really imagine users wanting to load an OpenAPI spec in a different way.

jerryjliu: yeah that's more intuitive, it unbundles loading the spec from actually calling the spec (with the requests tool). We can just treat the chatgpt plugin as a special case of the openapi spec

llama_hub/tools/library.json
```diff
@@ -41,6 +45,10 @@
     "id": "tools/slack",
     "author": "jerryjliu"
   },
+  "TextToImageToolSpec": {
```
ajhofmann, line 48: Tag along fix because I missed adding this in the initial text to image PR

jerryjliu (approved):

PR #407, opened 2023-07-20
Adding Dalle2 text to image tool

ajhofmann:


logan-markewich (approved):
Looks good to me, super neat tool!

llama_hub/tools/text_to_image/base.py
```diff
@@ -0,0 +1,88 @@
...
+import requests
+
+class TextToImageToolSpec(BaseToolSpec):
+    """Text to Image tool spec."""
+
+    spec_functions = ["generate_images", "show_images", "generate_image_variation"]
+
+    def __init__(self, api_key: Optional[str] = None) -> None:
+        if api_key:
+            openai.api_key = api_key
```
logan-markewich, line 16: One note: changing this under the hood might have unintended consequences, especially since it's a global singleton. I wonder if we can pass the api key into the API calls themselves?

ajhofmann: Yeah, I see what you mean. The only two ways documented on OpenAI's website for the python package seem to be environment variable or like the above code. Maybe just returning the auth error and having the user configure the env or package is ideal here.

logan-markewich: Seems like an undocumented feature 
https://github.com/openai/openai-python/blob/b82a3f7e4c462a8a10fa445193301a3cefef9a4a/openai/api_resources/image.py#L22

PR #415, opened 2023-07-23
Adding SEC Filings Loader

Athe-kunal:
This loader downloads the SEC filings documents like 10-K and 10-Q for multiple years and for multiple companies, and extracts the text portion using the Unstructured library. It also includes support for amended documents.


llama_hub/sec_filings/base.py
```diff
@@ -0,0 +1,87 @@
+from sec_filings import SECExtractor
```
jerryjliu, line 1: The default way to use llama_hub is now as a package, and so imports should include absolute paths to other files in the same directory: `from llama_hub.sec_filings.sec_filings import SECExtractor`

This will work for now. It won't work with the `download_loader` workflow, but if you want to support that you could do something like 
```python
try:
  from llama_hub.sec_filings.sec_filings import SECExtractor
except ImportError:
  # relative import from file
  from sec_filings import SECExtractor
```

jerryjliu (commented):
thanks for the submission! did an overall skim, comment on the imports.

what is `prepline_sec_filings` meant for? is this a demo app using the loader? would prefer to remove if not directly relevant to the loader

llama_hub/sec_filings/base.py
Athe-kunal: Hi @jerryjliu 
Thank you for your feedback
1. The `prepline_sec_filings` is the directory from Unstructured library SEC filings [link](https://github.com/Unstructured-IO/pipeline-sec-filings/tree/main/prepline_sec_filings). This library is useful to get the text from the filings. There are direct pypi package for the repo, hence I decided to just get the folder. I have referenced in the reference section.

Athe-kunal: I incorporated the changes that you asked for in the imports. Also, the `prepline_sec_filings` is essential for my loader as it is useful to get all the texts. As the unstructured library for SEC filings does not have a Pypi package, so importing the folder is useful.

jerryjliu: i see, thanks

jerryjliu (commented):
generally looks good. would you be able to add an entry to library.json? will help it show up in the UI

Athe-kunal:
Hi @jerryjliu 
I have added it to `library.json`. I will keep on improving this loader. Right now, it has only texts and it is missing out on tables. So now I am working on table integration. Thanks

jerryjliu (approved):

jerryjliu (changes requested):
wait you checekd in poetry.lock again

jerryjliu (approved):

PR #416, opened 2023-07-24
Adding personal access token (PAT) authentication in Jira reader

jinjintonic:
Jira reader uses misleading constructor argument name "api_token" but uses it as user password in basic auth. This change set renames email -> username, api_token -> password, as well as adds personal access token (PAT) option to be able to provide access to Jira instance without exposing user credentials.

### DIFF
--- a/llama_hub/jira/base.py
+++ b/llama_hub/jira/base.py
@@ -24,16 +24,38 @@ class JiraReader(BaseReader):
     """Jira reader. Reads data from Jira issues from passed query.
 
     Args:
-        email (str): Jira email.
-        api_token (str): Jira API token.
+        username (str): Jira username, requires password to be provided.
+        password (str): Jira password, requires username to be provided.
+        personal_access_token (str): Jira personal access token (PAT) value.
         server_url (str): Jira server url.
     """
 
-    def __init__(self, email: str, api_token: str, server_url: str) -> None:
+    def __init__(
+            self,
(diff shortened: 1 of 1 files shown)


llama_hub/jira/base.py
```diff
@@ -24,16 +24,38 @@ class JiraReader(BaseReader):
...
+        username (str): Jira username, requires password to be provided.
+        password (str): Jira password, requires username to be provided.
+        personal_access_token (str): Jira personal access token (PAT) value.
         server_url (str): Jira server url.
     """
 
-    def __init__(self, email: str, api_token: str, server_url: str) -> None:
+    def __init__(
+            self,
+            username: str = None,
```
jerryjliu, line 35: nit: typing should be `Optional[str] = None`

jerryjliu (commented):
looks good, one nit

EmanuelCampos:
bump @jinjintonic, can you fix the typing and optionals?

EmanuelCampos:
They accept username and password but accept email and the api_token as well

I would prefer to not give the option to use `username` and `password` since api_token is very easy to use

We can follow up with email, api_token, and add the personal access token as well (even not sure if it has a difference from api_token, I think not)

EmanuelCampos:
Closing this PR now due to inactivity as well, a merge with the main will be necessary, feel free to reopen

PR #424, opened 2023-07-25
Create a tool notebook

ajhofmann:
Fun little example of working with an Agent to automatically create typed function definitions and tool prompts from another python file using the AST library to extract the function info we want to wrap.


jerryjliu (commented):
is the next step to make the `PythonFileToolSpec` into a specific tool on llamahub?

jerryjliu (approved):

PR #428, opened 2023-07-27
Airbyte based readers

flash1293:
This PR adds 8 new readers:
* `AirbyteCDKReader` This reader can wrap and run all python-based Airbyte source connectors.
* Separate loaders for the most commonly used APIs:
  * `AirbyteGongReader`
  * `AirbyteHubspotReader`
  * `AirbyteSalesforceReader`
  * `AirbyteShopifyReader`
  * `AirbyteStripeReader`
  * `AirbyteTypeformReader`
  * `AirbyteZendeskSupportReader`

## Documentation and getting started
I added the basic shape of the config to the readme. This increases the maintenance effort a bit, but I think it's worth it to make sure people can get started quickly with these important connectors. This is also why I linked the spec and the documentation page in the readme as these two contain all the information to configure a source correctly (e.g. it won't suggest using oauth if that's avoidable even if the connector supports it).

## Document generation
The "documents" produced by these loaders won't have a text part (instead, all the record fields are put into the metadata). If a text is required by the use case, the caller needs to do custom transformation suitable for their use case.

## Incremental sync
All loaders support incremental syncs if the underlying streams support it. By storing the `last_state` from the reader instance away and passing it in when loading, it will only load updated records.


jerryjliu:
@flash1293 did you want us to review? let us know!

llama_hub/airbyte_zendesk_support/README.md
```diff
@@ -0,0 +1,56 @@
+# Airbyte ZendeskSupport Loader
+
+The Airbyte ZendeskSupport Loader allows you to access different ZendeskSupport objects.
+
+## Installation
+
+* Install llama_hub: `pip install llama_hub`
+* Install the zendesk_support source: `pip install source_zendesk_support`
```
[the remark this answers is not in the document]
pedroslopez: Looks like these are wrong in each one I've checked - worth looking through them.

Also, how does the requirements.txt come into play with these instructions?

llama_hub/airbyte_cdk/README.md
```diff
@@ -0,0 +1,42 @@
+# Airbyte CDK Loader
+
+The Airbyte CDK Loader is a shim for sources creating using the [Airbyte Python CDK](https://docs.airbyte.com/connector-development/cdk-python/). It allows you to load data from any Airbyte source into LlamaIndex.
+
+## Installation
+
+* Install llama_hub: `pip install llama_hub`
+* Install airbyte-cdk: `pip install airbyte-cdk`
+* Install a source via git (or implement your own): `pip install git+https://github.com/airbytehq/airbyte.git@master#egg=source_github&subdirectory=airbyte-integrations/connectors/source-github`
```
pedroslopez, line 9: (thought) wonder if we should gear these docs and this generic loader more as a way to use custom airbyte sources with llamaindex rather than installing our sources from github, since they should use our direct loaders instead for those.

pedroslopez (commented):
Looking good! Just a few comments, main thing is around the package names in the docs which don't seem to be correct.

llama_hub/airbyte_cdk/base.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+
+    def __init__(
+        self,
+        source_class: Any,
+        config: Mapping[str, Any],
+    ) -> None:
+        """Initialize with parameters."""
+
+        super().__init__(config=config, runner=CDKRunner(source=source_class(), name=source_class.__name__))
```
pedroslopez, line 27: Left the same comment on the langchain equivalent, but we could validate the config via spec at this stage here as well. Again, probably something to handle on the CDKRunner end.

llama_hub/airbyte_shopify/.gitignore
```diff
@@ -0,0 +1 @@
+test.py
```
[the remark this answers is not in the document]
flash1293: Just for local testing, I will remove

llama_hub/airbyte_cdk/base.py
```diff
@@ -0,0 +1,33 @@
...
+    
+    def _handle_record(self, record: AirbyteRecordMessage, id: Optional[str]) -> Document:
+        return Document(doc_id=id,text="", extra_info=record.data)
+
+    def load_data(self, *args: Any, **load_kwargs: Any) -> List[Document]:
+        return list(self._load_data(*args, **load_kwargs))
```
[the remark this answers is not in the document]
flash1293: Good question, what do you think, @jerryjliu ? Probably something we can split out of the first PR though

llama_hub/airbyte_salesforce/README.md
```diff
@@ -0,0 +1,59 @@
...
+  "is_sandbox": False, # set to True if you're using a sandbox environment
+  "streams_criteria": [ # Array of filters for salesforce objects that should be loadable
+    {"criteria": "exacts", "value": "Account"}, # Exact name of salesforce object
+    {"criteria": "starts with", "value": "Asset"}, # Prefix of the name
+    # Other allowed criteria: ends with, contains, starts not with, ends not with, not contains, not exacts
+  ],
+}
+```
+
+By default all fields are stored as metadata in the documents and the text is set to an empty string. Construct the text of the document by transforming the documents returned by the reader.
```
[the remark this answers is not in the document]
flash1293: I honestly thought of a simple for loop, added an example for that. I couldn't find a concept within llama index that would abstract this away, @jerryjliu what do you think about this?

llama_hub/airbyte_zendesk_support/README.md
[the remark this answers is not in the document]
flash1293: Fixed. There is a `download_loader` helper that looks at this requirements.txt and installs the requirements automatically.

llama_hub/airbyte_cdk/README.md
flash1293: De-emphasized this is a bit.

llama_hub/airbyte_cdk/base.py
flash1293: Yeah, I'll tackle this one on the CDK side

llama_hub/airbyte_hubspot/README.md
```diff
@@ -0,0 +1,58 @@
...
+    "credentials_title": "Private App Credentials",
+    "access_token": "<access token of your private app>"
+  }
+}
+```
+
+By default all fields are stored as metadata in the documents and the text is set to an empty string. Construct the text of the document by transforming the documents returned by the reader:
+```python
+for doc in documents:
+  doc.text = doc.extra_info["title"]
```
yisding, line 43: So to expand on the other comment further:

If I'm thinking about a hubspot use case, it may be let me do some Q&A over the previous notes and emails I've had with a partner.

The content of the notes and emails are probably what we want to be chunked and retrieved over. If we're only embedding each note individually without any additional text splitting, then we're OK because we embed all of the metadata by default.

However, if we want to start text splitting, then the problem arises, because 1. the splitting will only happen on doc.text, but 2. somewhat even more problematic, every split will have a complete copy of the document's metadata.

3000 words is a lot of text (definitely bigger than most interaction notes and emails) so we may be able to get away with it for now, but I think longer run we'll want a different document class to handle this kind of use case.

There could be a lot of interesting things here, because after the loader loads it in in this more structured way, the user or the application designer could say: I want my coworkers' Notion comments to be chunked alongside the content of the doc, or no, I don't want the comments to split alongside the doc and it's just a matter of adjusting a list for the fields to include in the splitting.

llama_hub/airbyte_cdk/base.py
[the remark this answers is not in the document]
jerryjliu: by default every loader implements `load_data` but you're free to define custom methods too!

llama_hub/airbyte_cdk/base.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any, List, Mapping, Optional
+
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+from airbyte_protocol.models.airbyte_protocol import AirbyteRecordMessage
+from airbyte_cdk.sources.embedded.base_integration import BaseEmbeddedIntegration
+from airbyte_cdk.sources.embedded.runner import CDKRunner
+
+
+class AirbyteCDKReader(BaseReader, BaseEmbeddedIntegration):
```
jerryjliu, line 10: would prefer composition vs. multiple inheritance, easier to inspect the specific object (BaseEmbeddedIntegration) you're using

llama_hub/airbyte_cdk/base.py
jerryjliu, line 33: out of curiosity where is _load_data defined?

llama_hub/airbyte_hubspot/README.md
jerryjliu: the other alternative is to dump *everything* into text actually (so naive text dump), and if the user wants to have more fine-grained metadata specifications they use this argument

jerryjliu:
this is awesome!!! 

also, can we add entries to library.json? will help it show up in the llamahub.ai UI

llama_hub/airbyte_hubspot/README.md
flash1293: I like the record handler plus dumping everything in there by default, seems like the easiest way to get started. The text could get really messy for some streams, but that's probably ok

llama_hub/airbyte_cdk/base.py
flash1293: it's on the base integration class that's coming from the CDK. We chose this approach so we can roll out improvements without having to update the loader itself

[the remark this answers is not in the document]
flash1293: OK cool, let's add a `lazy_load` as well that's returning an iterable.

jerryjliu: i see, makes sense

flash1293:
@jerryjliu Thanks for your patience, I addressed the following things:
* Composition over inheritance
* Added `lazy_load_data` which returns an iterator
* Added record handler to map records to documents
* Default mapping is putting everything into the text part now
* Add an example on how to use the record handler
* Added to library.json

llama_hub/airbyte_cdk/base.py
```diff
@@ -0,0 +1,54 @@
+import json
+from typing import Any, Callable, Iterator, List, Mapping, Optional
+
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+from airbyte_protocol.models.airbyte_protocol import AirbyteRecordMessage
```
jerryjliu, line 6: so we should figure out a better test system on our side, but currently tests are failing because we assume that third-party imports are lazy imports.

i know specifically for AirbyteRecordMessage it's used in the `RecordHandler` type used to type `record_handler`. For this do you think we could type `record_handler` as `Any` first, and then within `__init__` lazy import `AirbyteRecordMessage`, import `RecordHandler`, and do an `isinstance()` check on `record_handler`? 

A bit hacky but will at least allow tests to pass. thanks

jerryjliu (approved):
approving to unblock. one comment to get tests to pass. thanks for the help @flash1293 !

flash1293:
@jerryjliu Moved the imports into the constructor

PR #430, opened 2023-07-28
feat: minio loader

semoal:
Min.io is a high-performance object storage with a 100% compatible API with S3, that's open-source.
https://min.io

I've intregrated a loader to work similar as how it works the S3 loader, tested on my end with our code and works super good!


jerryjliu (approved):
this is great!

PR #431, opened 2023-07-29
Extending functions for DeepDoctectionReader

JaMe76:
Hi!

After spending quite a while extending deepdoctection to build a flexible data loader for PDFs and image scans I have taken the opportunity to extend the functionality of `DeepDoctectionReader`:

- You can build `Document`s by extracting the full text page-wise or by using layout sections like title, text, tables. 
- Adding metadata to `Document`s was already possible, I just made it a bit more clear, what type of metadata can be added.
- Customizing the loader. You can switch between three different OCR tools (Tesseract, DocTr, and AWS Textract), can do some filtering (e.g. excluding tables) and a lot more.

To give a little more insight about what is possible, I added some sections to the README. 

Finally, I changed the `requirements.txt`, so that users who just get started have a minimal setting and a low entry barrier.

### DIFF
--- a/llama_hub/file/deepdoctection/README.md
+++ b/llama_hub/file/deepdoctection/README.md
@@ -1,11 +1,14 @@
 # DeepDoctection Loader
 
-This loader extracts the text from a local PDF file using the deepdoctection Python package, a library that performs
-doc extraction and document layout.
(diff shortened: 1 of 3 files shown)


jerryjliu (approved):
thanks!

PR #439, opened 2023-08-02
Added Semantic scholar reader

shauryr:
This module loads papers from semantic scholar search for which the user can question and get answers to. It is using the [semanticscholar](https://pypi.org/project/semanticscholar/) library to search and get results. 

The user can define a broader search space, get the `Documents` loaded in `llama_index` and then ask questions as demoed here -

image


jerryjliu:
this is awesome! will review soon

llama_hub/semantic_scholar/README.md
```diff
@@ -0,0 +1,56 @@
...
+```python
+
+from llama_index.llms import OpenAI
+from llama_index.query_engine import CitationQueryEngine
+from llama_index import (
+    VectorStoreIndex,
+    ServiceContext,
+)
+
+s2reader = download_loader('SemanticScholarReader')
```
jerryjliu, line 22: nit: you never import `download_loader` from `llama_index`

also recommended way now is `from llama_hub.semantic_scholar.base import SemanticScholarReader` (e.g. import as package)

shauryr: ahh thanks for pointing that out. I have included it here 20819eae3bf2d7d67aeeaf6f3460e3aa7fe19e4f

jerryjliu (approved):

llama_hub/semanticscholar/base.py
```diff
@@ -0,0 +1,103 @@
+import logging
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+from semanticscholar import SemanticScholar
```
Disiok, line 4: could you do the import from the class constructor or methods? The reason is that we want this to be lazily imported. 

Take a look at other loader implementations as example!

shauryr: thank you for this suggestion! I have moved it here https://github.com/emptycrown/llama-hub/pull/439/commits/977671c4763adf24f307b4be55b9b34b667441b8

jerryjliu (approved):

PR #442, opened 2023-08-03
Adding Shopify GraphQL Tool Spec and Demo

ajhofmann:


llama_hub/tools/shopify/base.py
```diff
@@ -0,0 +1,36 @@
+"""Shopify tool spec."""
+
+from llama_index.tools.tool_spec.base import BaseToolSpec
+from typing import Optional
+import shopify
+
+class ShopifyToolSpec(BaseToolSpec):
```
jerryjliu, line 7: out of curiosity could we create a general "GraphQLAPIToolSpec" that could run graphql queries against any graphql http endpoint? 

to be clear we should def keep this one

jerryjliu (commented):
this demo is great

llama_hub/tools/shopify/base.py
ajhofmann: Will work on splitting out a GraphQLAPIToolSpec and SchemaDefinitionLanguageLoader today 👌

llama_hub/tools/shopify/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+In particular, the tool is very effective when combined with a method of retriving data from the GraphQL schema defintion.
+
+```python
+from llama_hub.tools.shopify.base import ShopifyToolSpec
+from llama_index.agent import OpenAIAgent
+
+tool_spec = ShopifyToolSpec('your-store.myshopify.com', '2023-04', 'your-api-key')
+
+agent = OpenAIAgent.from_tools(tool_spec.to_tool_list())
```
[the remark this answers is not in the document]
ajhofmann: Updated!

jerryjliu (approved):

PR #444, opened 2023-08-03
RSS Feed / OPML reader and article parser

ruze00:
Submitted to llama-index and tested.  PR here also in response to suggestion by @jerryjliu.


jerryjliu:
awesome thanks! will review soon

llama_hub/web/news/README.md
```diff
@@ -0,0 +1,52 @@
+# News Article Loader
+
+This loader makes use of the `newspaper3k` library to parse web page urls which have news
+articles in them.
+
+## Usage
+Pass in an array of individual page URLs:
+```python
+from llama_index import download_loader
```
jerryjliu, line 9: canonical way to use loaders is now `from llama_hub.web.news.base import NewsArticleReader`

llama_hub/web/news/base.py
```diff
@@ -0,0 +1,182 @@
...
+            if self.use_nlp:
+                metadata["keywords"] = getattr(article, "keywords", [])
+                metadata["summary"] = getattr(article, "summary", "")
+
+            documents.append(Document(text=content, metadata=metadata))
+
+        return documents
+
+
+class RssNewsReader(BaseReader):
```
jerryjliu, line 91: could you make this a separate reader with its own folder? will keep every entry distinct

jerryjliu (commented):
thanks - once you separate out both loaders, would you be able to add both to library.json?

llama_hub/web/news/base.py
ruze00: Sure.  There is already an `rss` folder.  Should I make it `rss_news` or other preference?

ruze00: updated in the meanwhile.

llama_hub/library.json
```diff
@@ -622,5 +622,23 @@
       "bucket",
       "storage"
     ]
+  },
+  "NewsArticleReader": {
+    "id": "news",
```
jerryjliu, line 627: id needs to be `web/news` and `web/rss_news` (to show up on llamahub UI)

jerryjliu (approved):
awesome! one final comment but approving to unblock

llama_hub/library.json
danipolo: this seems to be broken

PR #449, opened 2023-08-07
changed num to init for better declaration

rarisenpai:
I realized it's better to initialize through the init. so it would be 
google_spec = GoogleSearchToolSpec(key, engine_id, num)


jerryjliu (commented):
the main code changes look good 

could you not check in the poetry.lock file?

PR #459, opened 2023-08-10
Adding Azure speech, translate and CV tools

ajhofmann:
One question about this; adding the files to the `data` folder is nice for the demos, but we should make sure we aren't packaging the notebooks directory in our pip package, since the files are going to start making the download needlessly large.


jerryjliu:
> One question about this; adding the files to the `data` folder is nice for the demos, but we should make sure we aren't packaging the notebooks directory in our pip package, since the files are going to start making the download needlessly large.

regarding that point could we upload to dropbox and just `wget` it?

ajhofmann:
> > One question about this; adding the files to the `data` folder is nice for the demos, but we should make sure we aren't packaging the notebooks directory in our pip package, since the files are going to start making the download needlessly large.
> 
> regarding that point could we upload to dropbox and just `wget` it?

For this file I think we can actually just download it from the source, will look into the other files in the `data` folder to see whats worth refactoring into dropbox.

ajhofmann:
https://github.com/emptycrown/llama-hub/assets/10040285/beb34bcc-0b31-4042-a7a7-1725ecc86e5c

DV demo, working on a demo video for the speech tools

jerryjliu:
awesome!

jerryjliu (approved):
looks good! 

out of curiosity did you end up using dropbox?

PR #466, opened 2023-08-16
Adding Metaphor tool and notebook

ajhofmann:
Tool and guide to seamlessly connect an LLM to https://metaphor.systems/


llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,100 @@
...
+    """Metaphor tool spec."""
+
+    spec_functions = ["metaphor_search", "retrieve_documents", "current_date", "find_similar"]
+
+    def __init__(self, api_key: str) -> None:
+        """Initialize with parameters."""
+        from metaphor_python import Metaphor
+        self.client = Metaphor(api_key=api_key)
+
+    def metaphor_search(
```
jeffzwang, line 20: What's the reasoning behind prefixing `metaphor` here but not in the other functions? I assumed this is to help the agent better understand the tool by giving it a less generic name, but should that then have consistency across the 3 functions?

llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,100 @@
...
+        include_domains: Optional[List[str]] = None,
+        exclude_domains: Optional[List[str]] = None,
+        start_published_date: Optional[str] = None,
+        end_published_date: Optional[str] = None
+    ) -> str:
+        """
+        Metaphor allows you to use a natural language query to search the internet
+
+        Args:
+            query (str): A natural language query phrased as an answer for what the link provides, ie: "This is the latest news about space"
```
jeffzwang, line 33: This will need heavy prompt engineering in order for something like GPT3.5 to form proper queries. We have the proper prompts and could give them. Alternatively, the "autoprompt" flag can be used, and we use our own LLMs to form the query. Thoughts on what the better pattern would be? I'm a little worried about too lengthy of a prompt crowding the context window here - it would be about 500 tokens. Not sure what kind of context window budget considerations are at play.

llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,100 @@
...
+    spec_functions = ["metaphor_search", "retrieve_documents", "current_date", "find_similar"]
+
+    def __init__(self, api_key: str) -> None:
+        """Initialize with parameters."""
+        from metaphor_python import Metaphor
+        self.client = Metaphor(api_key=api_key)
+
+    def metaphor_search(
+        self,
+        query: str,
```
jeffzwang, line 22: The other useful field that I might put here is `type`, although that will complicate how the LLM perceives the tool. `type` can be `keyword` or `neural`, giving ultimate flexibility as to how to call Metaphor. But I guess there's a tradeoff with ease of use of the tool. "Keyword" search could also be its own separate tool. Maybe named `traditional_search` or something like that.

Thoughts?

llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Metaphor allows you to use a natural language query to search the internet
+
+        Args:
+            query (str): A natural language query phrased as an answer for what the link provides, ie: "This is the latest news about space"
+            num_results (Optional[int]): Number of results to return. Default 3.
+            include_domains (Optional[List(str)]): A list of top level domains like ["wsj.com"] to limit the search to specfic sites.
+            exclude_domains (Optional[List(str)]): Top level domains to exclude.
+            start_published_date (Optional[str]): A date string like "2020-06-15". Get the date from `current_date`
+            end_published_date (Optional[str]): End date string
```
jeffzwang, line 38: Just curious, is omitting the further detailing of what the field is because it saves tokens, and the LLM will already know what to do based on the previous example?

llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,100 @@
...
+        url: str,
+        num_results: Optional[int] = 3,
+        start_published_date: Optional[str] = None,
+        end_published_date: Optional[str] = None
+    ) -> str:
+        """
+        Retrieve a list of similar documents to a given url
+
+        Args:
+            url (str): The article to find similar articles to
```
jeffzwang, line 76: It's really any web page with text content. Doesn't have to be an article

jeffzwang (commented):
This looks mostly good to me! Thanks for doing this.

One thought on the content:
- It's possible that a better default num_results for LlamaIndex is higher if the use case is always going to be loading to a vector store. That depends on how your users are using you guys, I guess.

llama_hub/tools/metaphor/base.py
ajhofmann: good point, I think that just `search` is fine 👍

llama_hub/tools/metaphor/base.py
ajhofmann: I think it makes sense to enable autoprompt, no point in trying to prompt the LLM twice; will update 👍

llama_hub/tools/metaphor/base.py
ajhofmann: I'll play around with this a bit tomorrow and let you know what seems most effective

llama_hub/tools/metaphor/base.py
ajhofmann: Yeah exactly

llama_hub/tools/metaphor/base.py
ajhofmann: Gotcha, will rework some of the wording

llama_hub/tools/metaphor/base.py
ajhofmann: If we enable useAutoprompt, does the `type` field still have an effect?

jeffzwang: if `type` is `keyword` and autoprompt is on, autoprompt won't do anything. I wonder if a better thing to do would be to convert the input into a google-style query, but I imagine the LLM will already have done that

llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,104 @@
...
+            ids (List(str)): the ids of the documents to retrieve
+        """
+
+        response = self.client.get_contents(ids)
+        return [Document(text=con.extract) for con in response.contents]
+
+    def find_similar(
+        self,
+        url: str,
+        num_results: Optional[int] = 3,
```
jeffzwang, line 75: i see you changed the other value to 10. do you want this to be 10 too? personally i'm not sure of the merits of 3 vs 10. 3 may be better if there's no filtering being done

jeffzwang (commented):
Thanks for the edits - still have some high level questions that might inform how this is designed and how it could maximally ergonomic and useful for users.

1. Should we try to make this an omni-tool that supports Metaphor neural search as well as the keyword based search? That would make this an omni-tool for connecting to the Internet, making for a really easy developer experience.
2. There's the lingering question of how many results would be most useful to your users. 3? 10? 30? I'm not sure, and I imagine it depends on what users are doing.
3. How long could these docstrings be without impacting the performance of the whole system? Ideally we'd educate the LLM a good amount more on what Metaphor is but there is obviously a tradeoff.

llama_hub/tools/metaphor/base.py
jerryjliu: @jeffzwang did you imagine that type="keyword" or type="neural" would be set by the LLM or by the human? 

if set by the LLM, we would need to prompt it on best practices onwhen to use keyword or neural. 

otherwise we have two options from my understanding:
- use autoprompt=True
- let user manually set type in `__init__`

out of curiosity is the reason you don't like autoprompt=True because it's slow? that would be a fair consideation

llama_hub/tools/metaphor/base.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, api_key: str) -> None:
+        """Initialize with parameters."""
+        from metaphor_python import Metaphor
+
+        self.client = Metaphor(api_key=api_key)
+
+    def search(
+        self,
+        query: str,
+        num_results: Optional[int] = 10,
```
jerryjliu, line 30: @ajhofmann from your experiments do you think the LLM can infer a reasonable value? or does it usually leave it blank

jerryjliu (commented):
@jeffzwang thanks for your thoughts! at a high-level:

> * Should we try to make this an omni-tool that supports Metaphor neural search as well as the keyword based search? That would make this an omni-tool for connecting to the Internet, making for a really easy developer experience.

I agree that would be ideal. see below for comment on how this can be possible. if we don't use autoprompt and want LLM to infer we'd need best practices on using neural vs. keyword.

> * There's the lingering question of how many results would be most useful to your users. 3? 10? 30? I'm not sure, and I imagine it depends on what users are doing.

My take is we pick a reasonable default and go from there.

There is a high-level question @ajhofmann of how much of this the LLM can reasonably infer, vs. we should let the user customize somehow in the `__init__`. Don't think our tools support manual overrides by users atm.  

> * How long could these docstrings be without impacting the performance of the whole system? Ideally we'd educate the LLM a good amount more on what Metaphor is but there is obviously a tradeoff.

Good question! We haven't tested this extensively. could be good to experiment on in future iterations 

Seems like main action item woudl be enabling neural/keyword search. thoughts?

llama_hub/tools/metaphor/base.py
ajhofmann: I think the LLM is reasonable at using this field. For example, I changed the wording of `Can you summarize all of the news published in the last month on superconductors` to `some of the news` and it specified num_results = 5, with `all news` it uses the default of 10.

ajhofmann:
Thanks for the feedback everyone! 

> Should we try to make this an omni-tool that supports Metaphor neural search as well as the keyword based search? That would make this an omni-tool for connecting to the Internet, making for a really easy developer experience.
>> I agree that would be ideal. see below for comment on how this can be possible. if we don't use autoprompt and want LLM to infer we'd need best practices on using neural vs. keyword.

I think setting the keyword/autoprompt settings through init works. Otherwise we could expose two tools, one that uses autoprompt and one that uses keyword search, and let the agent know in the doc string the difference.

> There's the lingering question of how many results would be most useful to your users. 3? 10? 30? I'm not sure, and I imagine it depends on what users are doing.
>> My take is we pick a reasonable default and go from there.

Agreed, as I noted the LLM is pretty good at specifying the number of results if it needs to as well

> How long could these docstrings be without impacting the performance of the whole system? Ideally we'd educate the LLM a good amount more on what Metaphor is but there is obviously a tradeoff.
>> Good question! We haven't tested this extensively. could be good to experiment on in future iterations

Currently there is a limit on OpenAI's function calling API. The prompt for search is pretty close to the limit, and when you hit it we get `{docstring} is too long - 'functions.0.description'`

llama_hub/tools/metaphor/base.py
jeffzwang: It can be set by either the LLM or the human - if set by the LLM, it requires a ~150 token prompt. So not sure what the tradeoff is. We can handle it with a special search type that we have that can auto-decide between the two, with not _that_ much latency overhead.

But yes, autoprompt in general adds a good amount of latency - brings our search from the 250-300ms range to the 600-1000ms range. What do you think is the right tradeoff for LlamaIndex users?

ajhofmann:
> num_results default

Sure I think a sane default would be 5 to balance between context window/good number of results

> Auto prompt / Function calling

I think leaving autoprompt as the default is going to give us the best results, I think it's a good pattern to have the LLM behind the scenes improving the prompt, and it saves us tokens on the users side for better responses.

We have autoprompt enable by default right now, so is the only change we need to make to update the num_results default? 

I also did play around with adding a neural_search and keyword_search tool, but I think the amount of prompting to make it better than neural prompting is probably not worth it in terms of tokens, considering you have a model specifically focused on improving the prompts

llama_hub/tools/metaphor/README.md
```diff
@@ -0,0 +1,31 @@
...
+)
+agent = OpenAIAgent.from_tools(metaphor_tool.to_tool_list())
+
+agent.chat('Can you summarize the news published in the last month on superconductors')
+```
+
+`metaphor_search`: Search for a list of articles relating to a natural language query
+`retrieve_documents`: Retrieve a list of documents returned from `metaphor_search`.
+`find_similar`: Find similar documents to a given URL.
+`current_date`: Utility for the Agent to get todays date
```
jeffzwang, line 29: think we need to additionally document the new search and retrieve tool

jeffzwang (commented):
looks mostly good. there are some misc typos and gramamr things in the ipynb that I'd like to fix if you want to do a quick scan - hard to leave comments on the ipynb file. we can merge this maybe and then I could do a 15 min pass on that. or if you want to fix up that works too

llama_hub/tools/metaphor/README.md
ajhofmann: Added 👍

ajhofmann:
Took another pass at the grammar, I think it makes sense to merge this if you're good with it and we can fast follow on some more notebook fixes 👌

jerryjliu (approved):

PR #481, opened 2023-08-23
Adding MultiOn browsing tool

ajhofmann:
Adds a tool to connect the MultiOn browsing agent as a LlamaHub tool: https://www.multion.ai/


jerryjliu (approved):
@ajhofmann approving to unblock, though i have some comments:
- " to obtain the most recent blog post from openai" - this doesn't seem accurate
- in general what does adding an email tool do? I'm not actually i get the point of it vs. using multi-on directly. at a high-level i get the idea of "pre-seeding with context" but not sure how the example here highlights it
- is it possible to have a demo video of the nb in action? would help sell the integration

ajhofmann:
> @ajhofmann approving to unblock, though i have some comments:
> 
> * " to obtain the most recent blog post from openai" - this doesn't seem accurate

Oops yeah that was leftover from a previous iteration of this notebook, will fix that up.

> * in general what does adding an email tool do? I'm not actually i get the point of it vs. using multi-on directly. at a high-level i get the idea of "pre-seeding with context" but not sure how the example here highlights it

Yeah, I'll work on an example where seeding with the context makes a bit more sense 👌 

> * is it possible to have a demo video of the nb in action? would help sell the integration

💯

jerryjliu (approved):

PR #482, opened 2023-08-25
Enhancement to Semantic Scholar Loader - full text reader

shauryr:
Added document loading option for full text extracted from downloaded open access pdfs. 

As shown below, just marking `full_text` flag as `True` will download all the relevant open access pdfs and extract full text from the pdf and load it.

```python
documents = s2reader.load_data(query_space, total_papers, full_text=True)
```

- [x] Added demo notebook
- [x] Updated the README
- [x] Added pypdf and arxiv to `requirements.txt`

Attaching a screenshot of the loader in action! 

image


jerryjliu (approved):

jerryjliu:
sorry for the delay! merged, thanks @shauryr

PR #485, opened 2023-08-25
iterate json data to Document object

ron-unstructured:


jerryjliu (commented):
hey @ron-unstructured out of curiosity what's the context for the change? was it throwing a bug?

ron-unstructured:
Hi @jerryjliu, Thanks for the review. I discussed the current behavior of JSONLoader with Yi Ding. The current behavior is to load all the records into one Document. The proposed PR is to return a list of Documents, one each for the record. For example:
```
{
  { 'metadata': { 'source': <URL>, 'title': <page title> }, 'page_content': <text> },
  { 'metadata': { 'source': <URL>, 'title': <page title> }, 'page_content': <text> },
  { 'metadata': { 'source': <URL>, 'title': <page title> }, 'page_content': <text> },
}
```
This example will return a list of three Documents with the proposed PR, instead of one Document in the current behavior. Please let me know what you think. Thanks.

ron-unstructured:
> thanks! could this be an option that the user toggles?

Not in the proposed PR. But, that could be useful to provide as an option to toggle.

yisding:
Thanks for doing this Ron!

jerryjliu (approved):

godwin3737:
@jerryjliu , @ron-unstructured , this change has essentially broken JsonReader, also reported here https://github.com/run-llama/llama-hub/issues/606 

The issue seems to be:

1. Change was done to accommodate a schema that seems is not json syntax complaint.

json-syntax-incorrect

2. The change is not passing the actual value , but only the keys for syntactically accurate Json. Refer https://github.com/run-llama/llama-hub/issues/606 .

3. The line causing the issue : https://github.com/run-llama/llama-hub/blob/c36cdc54b82ced1bffe792293d896ca5681a2e61/llama_hub/file/json/base.py#L79

I will raise an issue with Toggle functionality.

godwin3737:
PR is created : https://github.com/run-llama/llama-hub/pull/816

PR #493, opened 2023-09-01
hwp_loader_init

sangwongenip:
This loader reads the HWP file, which is the format of many official documents in South Korea.


jerryjliu (commented):
awesome! mind adding some entries to library.json?

sangwongenip:
> awesome! mind adding some entries to library.json?

Done!

PR #500, opened 2023-09-03
json_entry_added_etc

sangwongenip:
1.Deleted 'hangeul' directory, whose contents are same with 'hwp'. Sorry for my mistake.
2.Added entry for extra info.

thanks!


jerryjliu (commented):
@sangwongenip I just realized, neither hangeul nor hwp were added to library.json right? could we add that? thanks

sangwongenip:
> @sangwongenip I just realized, neither hangeul nor hwp were added to library.json right? could we add that? thanks

Yep!

sangwongenip:
> @sangwongenip I just realized, neither hangeul nor hwp were added to library.json right? could we add that? thanks

I'll do it in 30 minutes!

jerryjliu:
thanks!

jerryjliu (approved):

PR #508, opened 2023-09-09
Elastic data reader

sadegh1404:
In this PR i've changed the elastic client to the official package for better usage and future development.

### DIFF
--- a/llama_hub/elasticsearch/base.py
+++ b/llama_hub/elasticsearch/base.py
@@ -4,11 +4,11 @@ This only uses the basic search api, so it will work with Elasticsearch and Open
 
 """
 
-
 from typing import List, Optional
 
 from llama_index.readers.base import BaseReader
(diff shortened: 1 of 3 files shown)


llama_hub/elasticsearch/base.py
```diff
@@ -4,11 +4,11 @@
 
 """
 
-
 from typing import List, Optional
 
 from llama_index.readers.base import BaseReader
 from llama_index.readers.schema.base import Document
+from elasticsearch import Elasticsearch
```
jerryjliu, line 11: let's do lazy imports (for tests to pass)

jerryjliu (commented):
looks good! one comment

llama_hub/elasticsearch/base.py
sadegh1404: It's done!

PR #512, opened 2023-09-13
Add reader for GitHub collaborators

rwood-97:
Add GitHub repository collaborators reader.
        
Retrieves the list of collaborators in a GitHub repository and converts them to documents.
Each collaborator is converted to a document by doing the following:
- The text of the document is the login _(thoughts on this welcome, I tried no text too but didn't work)._
- The title of the document is also the login.

The extra_info of the document is a dictionary with the following keys:
- login: str, the login of the user
- type: str, the type of user e.g. "User"
- site_admin: bool, whether the user has admin permissions
- role_name: str, e.g. "admin"              
- name: str, the name of the user, if available
- email: str, the email of the user, if available
- permissions: str, the permissions of the user, if available


llama_hub/github_repo_collaborators/base.py
```diff
@@ -0,0 +1,189 @@
...
+from importlib import metadata
+import logging
+import sys
+from typing import Dict, List, Optional, Tuple
+
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+
+if "pytest" in sys.modules:
+    from llama_hub.github_repo_collaborators.github_client import (
```
jerryjliu, line 31: is there a difference in the if / else condition? imports seem the same

jerryjliu (commented):
as a rough skim this mostly looks good! can we add to library.json?

llama_hub/github_repo_collaborators/base.py
rwood-97: Ah yes, will update.
Can also update in [github issues reader]https://github.com/rwood-97/llama-hub/blob/785bdfca30f30316e35db0b28e9892e7797c45bc/llama_hub/github_repo_issues/base.py#L30 and [github repo reader] https://github.com/rwood-97/llama-hub/blob/785bdfca30f30316e35db0b28e9892e7797c45bc/llama_hub/github_repo/base.py#L24 if you like - should I make a separate PR for these?

jerryjliu: sure! not sure how we didn't catch that before, thanks @rwood-97

jerryjliu (approved):

PR #525, opened 2023-09-19
fix: 🚑️ remove app id and secret from code

cestoliv:
There were some still valid credentials ! Here is an example of how to avoid hard-coding credentials, but the most important now is to revoke your credentials :


EmanuelCampos:
nice one

PR #544, opened 2023-09-27
break loop if API error

adayush:
# Description
When the API resulted in error, error got logged but the while loop continued infinitely.
Added a break to fix this.
https://github.com/emptycrown/llama-hub/issues/545

## Type of Change
- [x] Smaller change

## How Has This Been Tested?
Have tested this locally

## Suggested Checklist:
 I have performed a self-review of my own code
 New and existing unit tests pass locally with my changes


EmanuelCampos (commented):
Nice one

EmanuelCampos (approved):
Nice one

PR #552, opened 2023-09-29
Add JSONL functionality to JSONReader

sauravjoshi23:
# Description

With the growing adoption of JSONL for large datasets due to its line-delimited nature, there was a need to expand the capabilities of our existing JSONReader. Previously, the JSONReader was designed solely to handle standard JSON files. This PR addresses that limitation by adding functionality to JSONReader to handle JSONL files. This enhancement ensures that users can now work with both JSON and JSONL files without the need for any additional loaders or configurations. If required we can have separate loaders too to make them modular but the core logic for them remains the same.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Changes from llama_index/PR#7906 (#557)
 - fix: sql_wrapper utilities (#558)
 - Adding functionality for AsanaReader (#386)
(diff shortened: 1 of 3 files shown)


EmanuelCampos (commented):
nit: linting is broken can you fix that? `make format; make lint`

Thanks for the contribution

sauravjoshi23:
Sounds good Emanuel, will do it.

PR #556, opened 2023-10-02
Wordlift

msftwarelab:
# Description

Modifies the usage of the reader and methods of url handling.

# How Has This Been Tested?

I tested with the sample data and all worked fined.


llama_hub/wordlift/base.py
```diff
@@ -77,16 +81,20 @@ def fetch_data(self) -> dict:
...
         try:
             query = self.alter_query()
-            response = requests.post(self.endpoint, json={
-                "query": query}, headers=self.headers)
+            response = await asyncio.to_thread(
+                requests.post,
+                self.endpoint,
+                json={"query": query},
+                headers=self.headers,
+            )
```
EmanuelCampos, line 89: What's the reason to run the request on another thread? or add async operations?

EmanuelCampos (commented):
I did one comment, adding async not sounds a big deal. but I'm curious to understand better what you want to achieve

EmanuelCampos:
quick bump @msftwarelab

EmanuelCampos:
bump @msftwarelab

msftwarelab:
Sorry for late response @EmanuelCampos 
I used async for Multiple HTTP requests.
In the is_valid_html and clean_html functions, if the content is a URL, the function will make an HTTP request. This can lead to a lot of outgoing requests if many fields in the returned data from the WordLift GraphQL API are URLs. This can potentially cause the entire process to be stuck, especially if the URLs are non-responsive or slow.
Do you have any suggestions?

EmanuelCampos:
Thanks for the explanation @msftwarelab

I understand the problem, but we have some issues with eg:  [OnDemandLoader](https://docs.llamaindex.ai/en/stable/examples/tools/OnDemandLoaderTool.html )
runs synchronously

And probably other which I'm not able to find now but that needs to run synchronously

Maybe a solution would be to set a read timeout on the `is_valid_html` and `clean_html` functions, what do you think? https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

will cc: @logan-markewich on this one

msftwarelab:
Thank you for your kind response! @EmanuelCampos 
Ok. Let me try it.

EmanuelCampos:
quick bump @msftwarelab

PR #559, opened 2023-10-04
add url and timestamp to slack loader documents metadata

devinburnette:
# Description

Adds url and timestamp to slack loader documents so that you when you return source documents to a user they can get a link to the slack conversation with langchain and others.

## Type of Change

# How Has This Been Tested?

- [x] Tested locally

# Suggested Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - fix: sql_wrapper utilities (#558)
 - Adding functionality for AsanaReader (#386)
 - Add JSONL functionality to JSONReader (#552)
+- add url and timestamp to slack loader documents metadata (#559)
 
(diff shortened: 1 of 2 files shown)


EmanuelCampos (approved):

EmanuelCampos:
@devinburnette I noticed we can remove url from extra_info if isn't set, can you open another PR with this improvement?

EmanuelCampos:
> @EmanuelCampos so with the way I currently have it, if workspace_url isn't set, it just defaults to an empty string so URL still becomes `/archive/channel/timestamp` just without the base url prefixed, which still might be helpful to someone, but yea open to whatever you think is best.

Yea, i think it’s work well, don’t need a new PR, thanks

PR #572, opened 2023-10-07
added bitbucket loader

lejdiprifti:
# Description

In this PR, I have created a tool to load all the files inside a Bitbucket project that consists of multiple repositories as Documents to be used by an Index.

## Type of Change

- New Loader/Tool

# How Has This Been Tested?

I am applying the code in my company's bitbucket server and created an index using the documents loaded using the class I created.


EmanuelCampos (commented):
There's some issues on the code

image

Without the Authorization token, the requests probably will fail, can you adjust that?

Would suggest maybe a getHeaders function or something like that

And make sure to run `make format; make lint` please

llama_hub/library.json
```diff
@@ -983,5 +983,14 @@
       "repository",
       "collaborators"
     ]
+  },
+  "BitbucketReader":{
+    "id": "bitbucket_reader",
```
EmanuelCampos, line 988: nit: the id need to be the same as the folder name "bitbucket" to the tests pass

EmanuelCampos (commented):
awesome! did only one last comment, to the tests pass

llama_hub/library.json
lejdiprifti: thank you! just changed it

PR #579, opened 2023-10-10
Add AWS Athena Reader

matichon-vultureprime:
Reader for the AWS Athena.


llama_hub/athena/base.py
```diff
@@ -0,0 +1,74 @@
...
+    Instead, store them in environment variables or in a separate configuration file.
+
+    """
+
+    def __init__(
+        self,
+    ) -> None:
+        """Initialize with parameters."""
+
+    def create_engine(
```
EmanuelCampos, line 25: to avoid name shadowing, I think we should rename the function name to `create_athena_engine`, since we already import create_engine from sqlachemy

llama_hub/athena/base.py
```diff
@@ -0,0 +1,74 @@
+"""Athena Reader."""
+from typing import Optional
+import warnings
+import boto3
+from sqlalchemy.engine import create_engine
```
EmanuelCampos, line 5: you need to lazy import to the tests pass

PR #644, opened 2023-11-17
fix text splitter import

logan-markewich:


logan-markewich:
https://github.com/run-llama/llama_index/issues/8980

logan-markewich:
ugh need to update llama-index deps. One sec

PR #715, opened 2023-12-04
update llama2 dataset with baselines

jerryjliu:


nerdai:
oh shoot, sorry @jerryjliu. Going to close this as I updated your llama2 dataset with baselines already :)  Thanks!

PR #729, opened 2023-12-04
Added fix to run SQL query function description

wangdatan:
# Description

This change fixed an isseu which the run SQL function cannot correctly summarize the result. In addition to that, this includes a few refactorings to reuse code. And Fixed verbose logic, which correctly shows dataframe when run query with verbose=True

## Type of Change

- [ ] New Loader/Tool

# How Has This Been Tested?

- [ ] Added new notebook (that tests end-to-end)

# Suggested Checklist:

- [ ] New and existing unit tests pass locally with my changes


llama_hub/tools/waii/base.py
```diff
@@ -34,6 +34,32 @@ def __init__(
         WAII.Database.activate_connection(key=database_key)
         self.verbose = verbose
 
+    def _try_display(self, obj):
+        try:
+            # display df if it is available
+            if self.verbose:
+                display(obj)
+        except:
```
jerryjliu, line 42: why swallow the try/except here? is there an if/else statement instead?

jerryjliu (approved):
approving to unblock, other stuff lg

llama_hub/tools/waii/base.py
wangdatan: The reason is I only want to display the data frame with verbose=True. And if the tool logic runs outside of IPython, there's no `display` function. 

Alternatively, I can check if the function `display` exists or not, maybe that causes less confusion. I can also check if it is inside an IPython environment, but it seems more complex. e.g. https://discourse.jupyter.org/t/find-out-if-my-code-runs-inside-a-notebook-or-jupyter-lab/6935/7 

Let me know if you have any other ideas for this.

jerryjliu: ah i see. ok no problem

wangdatan:
@jerryjliu , I just updated the comments of the `try_display` method to make it less confusing. And I also updated the notebook, which includes the data frame output (tables) to make it easier to understand. And added a notebook block (see the last notebook) to do `chat('run it')` to make sure a query can be executed based on the context.

jerryjliu:
great! i think there's jsut a quick lint error. if you dont' have time to get to this i can do it in a few hours

wangdatan:
Oh, saw it, let me fix it very quickly!

wangdatan:
It is now fixed, @jerryjliu

PR #756, opened 2023-12-08
Update AzStorageBlobReader

JAlexMcGraw:
# Description

Added functionality to allow user to connect to blob storage with a connection string, instead of credentials and account URL. The credentials and URL functionality still exists. The motivation behind this was on a current work project, using the credentials and account URL did not function. Permissions weren't set right, but using the connection string, I could connect and I had one less parameter to specify.

Changed the temporary file names from random names to the names that are assigned in the container originally. The rationale is that sometimes the original names are needed later on, such as my case that the names were used to create metadata, which could not be done with random names. 

Updated README to reflect this new functionality, and added a tip on how to update file_extractor file to specify a specific different document reader, if needed. Updated library to reflect myself as a contributor/author to this module.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

Created an AzStorageBlobReader instance using only the container name and connection string for the storage account, and was able to load data using the .load_data() method.

To test yourself, instead of specifying a container name, credentials, and account URL, can specify just the container name and connection string. Then, run the load_data() method on the class instantiation.

- [ ] Added new notebook (that tests end-to-end) (Tested this locally and it worked, but did not upload. Would you like me to upload?)

# Suggested Checklist:

- [ ] I have performed a self-review of my own code
- [ ] New and existing unit tests pass locally with my changes (did not test)

### DIFF
--- a/llama_hub/azstorage_blob/README.md
+++ b/llama_hub/azstorage_blob/README.md
@@ -2,20 +2,44 @@
 
 This loader parses any file stored as an Azure Storage blob or the entire container (with an optional prefix / attribute filter) if no particular file is specified. When initializing `AzStorageBlobReader`, you may pass in your account url with a SAS token or crdentials to authenticate.
 
-All files are temporarily downloaded locally and subsequently parsed with `SimpleDirectoryReader`. Hence, you may also specify a custom `file_extractor`, relying on any of the loaders in this library (or your own)!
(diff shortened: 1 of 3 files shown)


anoopshrma (approved):

anoopshrma:
Awesome! Thanks for the contribution! 
All looks good for merge except linting, Could you take a look at that!!

JAlexMcGraw:
Fixed linting. Should be good to go!

nerdai:
Thanks @JAlexMcGraw for the contribution. FYI, I had to remove your name in `library.json` as entering a list of authors at current moment breaks the UI.

In lieu of not being listed as an author, I added an UPDATES section in the README indicating what updates you have made to this pack.

(cc @anoopshrma for future reference :) )

anoopshrma:
Ah I see, Thanks for the heads up @nerdai  🙌 
will keep this in mind

PR #769, opened 2023-12-13
Logan/dense x retrieval

logan-markewich:
# Description

Adds a llamapack with an initial implementation of the paper [Dense X Retrieval: What Retreival Granularity Should We Use?](https://arxiv.org/abs/2312.06648).


logan-markewich:
In general, my conclusion on this paper/method is that
a) it's very similar to retrieving based on KG triplets (just like our KG index)
b) its quite slow, expensive, using the LLM to generate propositions (similar to the KG index lol)

jerryjliu (approved):

jerryjliu:
wait @logan-markewich can you change the name of the notebook

jerryjliu:
it says "fuzzy citation query engine"

PR #773, opened 2023-12-14
Adding PreprocessReader for LlamaHub

preprocess-co:
# Description

Adding a new reader called `Preprocess` to interact with the `Preprocess API library`. 
`Preprocess` is a document chunking service for LLM, it splits any kind of document into chunks following the original document layout and content semantics. `Preprocess` can handle the following file types:
- PDF
- Office (Word, PowerPoint, and Excel)
- OpenOffice (.odt, .pds, and .odp)
- HTML (web pages, articles, and emails) and plain text

We have built `Preprocess` because standard basic chunking methods (sentence-based or on a fixed word count) impact the search accuracy and may create Garbage in/Garbage out situations.
If you decide to use `PreprocessReader` it is required to install `pypreprocess` library via `pip` command.
And contact [Preprocess](mailto:support@preprocess.co) for requesting an API Key.
## Type of Change

- [x] New Loader/Tool

# How Has This Been Tested?

`Preprocess` API library is a live service already integrated into applications. We recently released our Python SDK and then wrapped it to make it suitable for LlamaHub.

In tests folder you can find a `test_preprocess` folder that contains the 2 methods for testing and a pdf file for testing, each method is for testing each scenario as end-to-end, so first one for `load_data` which is just return the text of the file inside a Document object, and another mechond `get_nodes` which is consider the chunks from Preprocess and handle it inside a node and mark its next and previuos one.

- [x] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it made sense

# Suggested Checklist:

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works


pyproject.toml
```diff
@@ -21,6 +21,7 @@ html2text = "*"
 psutil = "*"
 retrying = "*"
 pyaml = "^23.9.7"
+pypreprocess = "^1.1.0"
```
nerdai, line 24: Do we need to add this as a project dependency? Should be fine to keep it in the requirements.txt as you have now, since these will be installed when someone installs this llama-pack

nerdai (commented):
This looks great! Thanks @preprocess-co for the contribution. One requested edit is that we don't add the package to the main project's list of deps.

pyproject.toml
preprocess-co: i think no, you've to add it at the final step, cause i thought that it's enough for requirements.txt but 2 of checks has been failed due to the auto dependency download i think relay on the poetry install only, after adding it all checks has been passed.
anyway i've kept it in requirements.txt so save the project structure as all others doing that, but to pass the checks it has to be in the project dependency.

jerryjliu (approved):

llama_hub/preprocess/base.py
```diff
@@ -0,0 +1,145 @@
+"""Preprocess Reader."""
+import os
+import hashlib
+from typing import List
+from llama_index.readers.base import BaseReader
+from llama_index.readers.schema.base import Document
+from llama_index.schema import TextNode, NodeRelationship, RelatedNodeInfo
+
+from pypreprocess import Preprocess
```
nerdai, line 9: typically we move this import under the __init__ method of the Reader. See for example: https://github.com/run-llama/llama-hub/blob/68fed88d8e36223c6d8fa50398b347add84cd66f/llama_hub/confluence/base.py#L41

nerdai: Can you kindly do the same?

nerdai: its okay @preprocess-co  — I took care of it. All checks pass now. Thanks for the contribution! merging now :)

preprocess-co: @nerdai Thank you, sorry i didn't notice it on time to handle it :)

PR #788, opened 2023-12-19
Improved s3 reader: Updated s3Reader so that users have more control over where they want to download their s3 files - randomly generated temp directory or a user-defined subdirectory under /tmp. Introduced a new paratmeter custom_temp_subdir in load_data() method

utility-aagrawal:
Updated s3Reader so that users have more control over where they want to download their s3 files - randomly generated temp directory (current behavior) or a user-defined subdirectory under /tmp. Introduced a new paratmeter custom_temp_subdir in load_data() method.

**Existing features will work as-is.**

This is just an additional feature and extends the application of this loader:

1) I implemented this because I wanted to make use of docstore upserts as explained here: https://docs.llamaindex.ai/en/stable/examples/ingestion/redis_ingestion_pipeline.html#run-the-redis-based-ingestion-pipeline
Current loader doesn't work with this since there's a random component in downloaded filepath which changes every time even if we are indexing the same file. This new feature makes it possible to do upserts effectively since we have removed the random component from the filepath

2) This feature will also decrease the storage requirement for ingestion cache and document store. Since the existing loader download files in a randomly generated temp subdirectory, cache and document store store duplicate information if everything else is same but the download filepath between subsequent runs of indexing.

3) New parameter custom_temp_subdir can be hardcoded or customized based on users' requirements. For example, in my use case, I am using a unique folder id to download files related to a particular folder. If users are downloading files from an s3 bucket, they can use their bucket name as custom_temp_subdir

Let me know if you need anything additional. Thanks!


nerdai:
Hey @utility-aagrawal, do you mind resolving conflicts. It looks like main has moved on you a little bit :)

utility-aagrawal:
@nerdai , resolved conflicts. Please review and let me know in case of any further questions. Thanks!

utility-aagrawal:
@nerdai , can you please review and merge this PR, if everything looks good? Thanks!

nerdai:
Sorry for the delay @utility-aagrawal looks great! Merging now.

utility-aagrawal:
Thanks a lot, @nerdai !

PR #790, opened 2023-12-19
Update PDFReader

JAlexMcGraw:
# Description

Added the ability to use PDFPlumber as the underlying library extracting text, because from my own testing it performed better than PyPDF, albeit a little slower. A user will need to install the PDFPlumber library to use this. The reason behind including this functionality is because it was able to extract more/much higher quality text from PDFs than the PyPDF implementation.

This would also now match the PDFReader from llama-index.

Fixes # (issue)

## Type of Change

- [x] New Loader/Tool
- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] Tested Locally

# Suggested Checklist:


anoopshrma:
Rather than changing the base PDF reader, I would suggest you create a new PDF reader using `PDFPlumber`

JAlexMcGraw:
Sounds good, can do!

PR #804, opened 2023-12-22
Extract metadata from Azure BLOB

raccoonex:
# Description

Extract useful system-generated metadata from Azure BLOB and user-defined metadata and tags.

The changes from this PR extract metadata from Azure BLOB entities. Before this PR the `llama_index.readers.file.base.default_file_metadata_func` was used for metadata extraction. However, since the data is downloaded to the host system, the extracted metadata may not be correct.

This PR implements metadata extraction directly from Azure BLOB properties which consists of system metadata (e.g. `creation_time`) and user-defined metadata (as `metadata` and `tags`).

The new metadata set equals to the one obtained by the default metadata extractor + other Azure system meta + user-defined meta.

No dependencies have been changed.

## Type of Change

# How Has This Been Tested?

This has been tested with Azure Storage Blob (authenticated with connection string) and a Python script.
Example:

```python
from llama_hub.azstorage_blob import AzStorageBlobReader

reader = AzStorageBlobReader(container_name="the-container-name", connection_string="conn-str")
documents = reader.load_data()

documents[0].metadata
{'page_label': '1', 'file_name': 'some_file_name.pdf', 'file_type': 'application/pdf', 'file_size': 1093814, 'creation_date': '2023-12-20', 'last_modified_date': '2023-12-21', 'last_accessed_date': None, 'container': 'the-container-name', 'source_url': 'this-is-user-defined-meta'}
```

# Suggested Checklist:

### DIFF
--- a/llama_hub/azstorage_blob/base.py
+++ b/llama_hub/azstorage_blob/base.py
@@ -83,15 +83,18 @@ class AzStorageBlobReader(BaseReader):
                 self.account_url, self.container_name, credential=self.credential
             )
         total_download_start_time = time.time()
+        blob_meta = {}
 
         with tempfile.TemporaryDirectory() as temp_dir:
             if self.blob:
-                stream = container_client.download_blob(self.blob)
+                blob_client = container_client.get_blob_client(self.blob)
+                stream = blob_client.download_blob()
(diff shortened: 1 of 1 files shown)


nerdai (approved):
Looks great! Thanks @raccoonex for this update!

PR #809, opened 2023-12-22
chore: get the full detailed text from Confluence

pesarkhobeee:
The idea is to get all the possible data from Confluence: 
https://developer.atlassian.com/server/confluence/expansions-in-the-rest-api/

For example, if you use the regular storage endpoint it will give you the ID of the mentioned users inside the wiki pages but in the export_view you will get their full name and family, this will apply to all data that needs some post process like data in code views. 

We tested this and we are super happy using export_view instead of storage end point.

### DIFF
--- a/llama_hub/confluence/base.py
+++ b/llama_hub/confluence/base.py
@@ -268,7 +268,12 @@ class ConfluenceReader(BaseReader):
         return ret
 
     def _get_cql_data_with_paging(
-        self, cql, start=0, cursor=None, max_num_results=50, expand="body.export_view.value"
+        self,
+        cql,
+        start=0,
+        cursor=None,
+        max_num_results=50,
+        expand="body.export_view.value",
     ):
         max_num_remaining = max_num_results
         ret = []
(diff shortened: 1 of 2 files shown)


anoopshrma:
Hi @pesarkhobeee , Thank you for the contribution!! 
I think you'll need to update the test case and take a look at linting too.

pesarkhobeee:
Hey @anoopshrma , sorry I didn't check the tests, now they are aligned with my changes. I should emphasize again on the changes, you can see the official documentation about the new endpoint:

> [EXPORT_VIEW](https://docs.atlassian.com/atlassian-confluence/6.5.2/com/atlassian/confluence/api/model/content/ContentRepresentation.html#EXPORT_VIEW)
HTML representation for viewing, but with absolute urls, instead of relative urls in the markup.

and compare it with the old one:

>[STORAGE](https://docs.atlassian.com/atlassian-confluence/6.5.2/com/atlassian/confluence/api/model/content/ContentRepresentation.html#STORAGE)
Raw database format, for content that stores data in our XML storage format

anoopshrma (approved):

anoopshrma:
Hey! 
Thank you for the contribution. Much appreciated!!

PR #827, opened 2023-12-30
feat: Add full site BFS scraping loader

an-bluecat:
# Description

Currently, there is no loader to scrape the full site. Instead, users need to scrape all the links themselves. With this loader, users can specify an entry point and a prefix to scrape and store all texts within a site.

## Type of Change
- [x] New Loader/Tool

# How Has This Been Tested?
tried modifying the download repo link and tried downloading it and tested basic functionality:
image

# Suggested Checklist:

### DIFF
--- a/llama_hub/web/whole_site/base.py
+++ b/llama_hub/web/whole_site/base.py
@@ -59,7 +59,6 @@ class WholeSiteReader(BaseReader):
         self.driver = self.setup_driver()
 
     def extract_content(self):
-
         WebDriverWait(self.driver, 10).until(
             EC.presence_of_element_located((By.TAG_NAME, "body"))
         )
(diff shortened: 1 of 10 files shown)


llama_hub/web/whole_site/base.py
```diff
@@ -0,0 +1,129 @@
+from typing import List
+from selenium import webdriver
+from selenium.webdriver.common.by import By
+from selenium.common.exceptions import WebDriverException
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+import chromedriver_autoinstaller
```
nerdai, line 7: Can we lazily import these, so that we don't need to make additions to the `test_requirements`?

As an example:

https://github.com/run-llama/llama-hub/blob/364feb5b9c41a081da53aae02e894c6caa2f478d/llama_hub/llama_packs/gradio_agent_chat/base.py#L36

nerdai (approved):
Thanks for the contribution @an-bluecat! Just have a somewhat minor requested change to what you've got here.

llama_hub/web/whole_site/base.py
an-bluecat: We can do that for chromedriver_autoinstaller, but for selenium, it might be a bit slow if we lazy import at different places of the functions. What do you think?

PR #834, opened 2024-01-03
Handle bytestrings in PDF reader

PhilippeMoussalli:
# Description

A fix for the PDFReader to handle reading files as bytestrings. The loader already supported that function but there was a small bug related to the fact that it expected a file name to be fetched which is not available when passing a pdf bystring. This PR addresses this. The file name can still be added manually using the `extra_info` argument.

PDF bytestrings can be useful in case the user wants to load in case a user uses a custom filesystem interface like fsspec

example:
```python
# Create the GCS file system
gcs = fsspec.filesystem('gcs')

# Open and read the PDF file from GCS
with gcs.open(f'{bucket_name}/{file_path}', 'rb') as file:
    pdf_content = BytesIO(file.read())
```

Fixes # (issue)

## Type of Change

- [ ] New Loader/Tool
- [X] Bug fix / Smaller change

# How Has This Been Tested?

I ran the class locally with different file inputs (path to pdf, pdf bytestrings)

# Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] New and existing unit tests pass locally with my changes
- [X] I ran `make format; make lint` to appease the lint gods

### DIFF
--- a/llama_hub/file/pdf/README.md
+++ b/llama_hub/file/pdf/README.md
@@ -4,7 +4,7 @@ This loader extracts the text from a local PDF file using the `PyPDF2` Python pa
 
 ## Usage
 
-To use this loader, you need to pass in a `Path` to a local file.
+To use this loader, you need to pass in a `Path` to a local file or a PDF byte stream.
 
 ```python
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


nerdai (approved):
thanks for adding logic for this use case @PhilippeMoussalli!

PR #915, opened 2024-01-31
Ionic shopping tool

stewartjarod:
# Description

Ionic is a shopping assistant tool for llms. I've added the tool, which depends on our Ionic API SDK library.

While there, I updated the typing packages.

**Note**: _I had issues with running tests locally, even on main, so I'm not entirely sure if this is good to merge without a CI/CD step for tests._ 

Fixes # (issue)

## Type of Change

- [x] New Loader/Tool

# How Has This Been Tested?

- [x] Added new notebook (that tests end-to-end)

# Suggested Checklist:

- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works


jerryjliu:
out of curiosity @stewartjarod what were the issues you ran into in running tests?

pyproject.toml
```diff
@@ -27,8 +27,8 @@ pytest = "7.2.1"
 pytest-dotenv = "0.5.2"
 pytest_httpserver = "1.0.8"
 pytest-mock = "3.11.1"
-typing-inspect = "0.8.0"
-typing_extensions = "^4.5.0"
+typing-inspect = "0.9.0"
```
jerryjliu, line 30: why this change

jerryjliu (commented):
this is a cool idea! 

to make sure, the main capability the API offers is query right? not actually being able to make a purchase decision yet?

stewartjarod:
I wasn't able to get the async tests to pass.

Right, at this time we provide a query for products and recommendations. Looking to expand from there. 😀

pyproject.toml
stewartjarod: I have the understanding with these Python types that it is good to keep them up to date and that they are entirely backward compatible. I could be wrong! I come from mostly a TS background and try to use types for everything.

stewartjarod:
Ah, that's a weird failure. While I'm figuring this out, @jerryjliu do you want me to revert the package update for typings?

pyproject.toml
stewartjarod: I can roll this back if it will cause issues

stewartjarod:
Tests should pass now 👍

stewartjarod:
Ah, that's a new one. Type responses like that must not be supported in 3.8. Sorry for the issues here with tests. Adding future annotations seemed to do the trick here.

jerryjliu (approved):

